PR #2, opened 2019-07-22
End reaction fix + option for orienting beams

FJFranklin:
I'm still trying to get my head around the maths of it, but:
1. A mechanism for orientating beams by specifying a reference point in the beam's ZX-plane.
2. Don't create the transformation matrix, T(), more than once.
3. I think T() needs to be transposed when transforming back to the global coordinate frame.

### DIFF
--- a/Examples/Double Beam - Point Load.py
+++ b/Examples/Double Beam - Point Load.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sun Sep 16 14:43:31 2018
+
+@author: craig
+"""
+# Example of a simply supported beam with a point load.
+# Units used in this example are metric (SI)
+
+import numpy as np
+
+# Import `FEModel3D` from `PyNite`
+from PyNite import FEModel3D
+
+# 1: No beam-reference point
+# 2: Ref @ [0,0,-10] - should match Case 1
+# 3: Ref @ [0,-10,0] - twists the beam through 90 degrees
+
+# *** Puzzle: Why do the node reactions flip upside down in Case 3 only? ***
+# (Actually, they go to zero for [0,-10,10].)
+
+case = 3
+
+if case == 1:
+    beamref = None
+if case == 2:
+    beamref = np.asarray([0,0,-10])
+if case == 3:
+    beamref = np.asarray([0,-10,0])
+
+# Create a new finite element model
+SimpleBeam = FEModel3D()
+
+# Add nodes (3 metres apart on x-axis)
+L = 3
+SimpleBeam.AddNode("N1", -L/2, 0, 0)
+SimpleBeam.AddNode("N2",  0,   0, 0)
+SimpleBeam.AddNode("N3",  L/2, 0, 0)
+
+# Add a beam with a solid rectangular section:
+E = 209E9  # Young's elastic modulus
+G = 79.3E9 # shear modulus
+b = 30E-2  # breadth (y-thickness)
+d = 1E-2   # depth   (z-thickness)
+Iyy = b * d**3 / 12 # 2nd moment of area
+Izz = b**3 * d / 12 # 2nd moment of area
+J = Iyy + Izz       # 2nd polar moment of area
+A = b * d           # cross-sectional area
+SimpleBeam.AddMember("M1", "N1", "N2", E, G, Iyy, Izz, J, A, beamref)
+SimpleBeam.AddMember("M2", "N2", "N3", E, G, Iyy, Izz, J, A, beamref)
+
+# Provide simple supports - can't displace or twist at either end
+SimpleBeam.DefineSupport("N1", True, True, True, True, False, False)
+SimpleBeam.DefineSupport("N3", True, True, True, True, False, False)
+
+# Add a point load at the midspan of the beam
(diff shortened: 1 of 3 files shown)


JWock82:
I've fixed the [T] matrix issue (I think). Let me know if you're still having problems. I'll look into your other comments/changes perhaps this weekend.

JWock82:
I've double check member stiffness matrix, transformation matrix, and nodal load vector creation and I'm pretty confident PyNite is correct.

Looking closer at the "Odd Behavior Noted" example you sent, I think you're confusing Iyy and Izz. You've defined Iyy as b*d**3/12, but unless the beam is laying on its side that's Izz, not Iyy. I ran it the other way and got the values I expected.

Iyy resists loads applied in the local z-direction. Izz resists loads applied in the local y-direction. This is consistent with how many commercial packages work.

I'll look into the reactions flipped upside down in Case 3 when I have some more time.

FJFranklin:
(Back from holiday now.)
You're right. I was confusing Iyy & Izz - thanks.
I was also confusing myself about how the axes were orientated.

The case thing is an attempt to make the local beam coordinate system controllable by providing a reference point to anchor the zx-plane. In Member3D.T() you construct a local coordinate system in a way that seems to ignore the extra degree of freedom.

I need to have another proper look at this.

Best wishes,
Frank

JWock82:
I've double checked, and I've constructed the T matrix the way my finite element textbook says to. It says that if the member's local x-axis is parallel to the global z-axis, the T matrix needs adjustment to locate the local y-axis properly. I've based PyNite off of "A first Course in the Finite Element Method, 4th Edition" by Daryl L. Logan.

FJFranklin:
Okay, so I've created a new fork based on your latest, and made a number of changes:
 - Defer matplotlib initialisation until actually needed.
 - Add a MemberBasis class (with Member3D as a subclass) to handle local coordinate system.
 - Add mechanism for controlling axial orientation of member.
 - Add Material and Section classes for more flexible handling of section properties.
 - Add wireframe drawing in 3D of members using VisPy - with deferred import.
 - Add double-beam example with various options for testing.

There was nothing wrong with your T-matrix construction, except that it ignores a degree of freedom - in an intelligent way, but there are cases where the axial orientation of the section needs to be controlled.

The deferred initialisation, by the way, is so important. At least on OSX, if two programs try to use matplotlib at the same time it causes issues (i.e., crashes), and just importing it does the damage. And by deferring the import of packages, they don't even need to be installed unless actually wanted.

Regards,
Frank

JWock82:
I'm still unsure of the intent of the MemberBasis class. The local coordinate system for the member never gets lost in the process of building the global stiffness matrix. Shear, moment, axial, and deflection results are available for each member along its local x-axis once calculations have been run. The local matrices are also still available. I'll have to look at your code a little closer to make sure I'm not missing something here.

I'll have a look at deferred initialization a little closer too. I'm interested in having a look into VisPy as well and seeing how you implemented it. Thanks for your help. I need to find some time to sit down and study your code closely.

FJFranklin:
I've been learning (and unlearning) so much about stresses these past few weeks - lots of fun, but also a pain in the neck. Even the bits I've done so far need to be validated, and I doubt it will ever be more accurate than surface stress +/-10%. (I've been using PyNite within my DomeBuilder optimisation program, and would like to extend that work a little, but need to more confident about the stress analysis side of it.)

In retrospect, it probably wasn't necessary to create a separate MemberBasis class, but it's fairly clean. The Double Beam example has a large number of options for exploring the stuff I've been adding in, so that's probably the best place to look.

JWock82:
You should be able to compute flexural stresses from the section modulus, M/S = M/(I/y). Where "I" is the moment of inertia about the axis being considered, and y is the distance from the neutral axis to the extreme tension or compression fiber. You can directly add uniform axial stresses, P/A, to these bending stresses for most materials. Shear stresses are V/A, except that flanges of members should usually be ignored when calculating the areas. Torsional stresses are usually not present if the member is detailed and loaded properly. The first rule of torsion is to eliminate it.

Most modern structural codes governing designs of materials take a strength design approach and are more interested in force effects rather than stresses.

When you say +/- 10% it sounds to me like you're trying to apply a failure criterion/envelope such as the Von Mises or Tresca criterion. Those types of criterion are generally not used in structural design. They only look at maximum distortion energy and ignore the potential for buckling mechanisms to develop.

FJFranklin:
Axial and bending stresses are, as you say, straightforward. The shear stresses are where it gets interesting... I suppose I'm approaching this from the perspective of someone who teaches students about stresses, but who has also shied away from the complication of shear stresses in bending.

By the way, I notice you're now using self.__plt to defer import. While it probably doesn't make much difference, it would make more sense to have this as a class variable rather than an instance variable. At the moment, each individual member that you plot will attempt the import.

Hmm... a p-delta engine. There's a thought. My PhD student has been analysing fuel storage tanks subject to wind and snow loads, but dynamic / earthquake effects are still a big TODO.

JWock82:
I think I've got that fixed now. I didn't realize there was such a thing as a class variable that applied to all instances of a class. That's a handy trick. I've still got a lot to learn about programming. Thanks for the help.

PR #3, opened 2019-07-22
Revert "End reaction fix + option for orienting beams"

JWock82:
Reverts JWock82/PyNite#2

After further review, I think the transformation matrix needs to be inverted rather than transposed. I'll look into this further and correct it accordingly.


FJFranklin:
For an orthonormal basis, is there a difference?

JWock82:
I'm not sure. I'm not an expert on linear algebra. You're probably right, but I'm going invert it just to be consistent with the textbooks I've used to built the rest of this project.

Also, up until today I've been keeping my master copy on my local computer. I've deleted and replaced all the files with newer versions today, which may have dissociated your pull request from the files. I did this so that GitHub has all my latest files. Going forward I'm going to clone this repository to my local machine and keep the master on GitHub.

I'll have a look over all your code and try to get it implemented into the master as soon as I can. I do this in my free time so it probably won't happen overnight, but I'll get on it.

FJFranklin:
Don't worry. I understand how life gets in the way of stuff like this. I'm really grateful that PyNite exists and was so easy to use. Take a look at my DomeBuilder repository if you want to see it in action.

PR #23, opened 2019-12-31
Update BeamSegment.py

LcnGnd:
Implementation of concentrated torque

### DIFF
--- a/PyNite/BeamSegment.py
+++ b/PyNite/BeamSegment.py
@@ -38,8 +38,6 @@ class BeamSegment():
       The displacement at the start of the segment
     EI : number
       The flexural stiffness of the segment
-    GJ : number
-      The torsional stiffness of the segment
     
     Methods
     -------
@@ -76,7 +74,6 @@ class BeamSegment():
         self.theta1 = None # Slope at start of beam segment
         self.delta1 = None # Displacement at start of beam segment
         self.EI = None # Flexural stiffness of the beam segment
-        self.GJ = None # Torsional stiffness of the beam segment
 
 #%%
     # Returns the length of the segment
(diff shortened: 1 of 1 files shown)


LcnGnd:
I'm not sure of semplification done in MaxTorsion and MinTorsion methods. How do you PyNiteFEA to distinguish the maximum and the minimum torsion if the two functions have the same instructions (return self.T1)?

LcnGnd:
Another question about sign convention of torsional moment. According with your choosing, torsion is positive counter-clockwise on the right side of a section cut. So, thinking about vector, how it is directed on the right side of a section cut?

JWock82:
Until distributed torques are incorporated, the torsion will always be constant on each beam segment. A constant function has no max or min. The MaxTorsion and MinTorsion functions are "dummy" functions for now. This is true for a beam segment, but not true for the overall beam. The beam MaxTorsion and MinTorsion functions will check each segment and report the max and min for all the segments in the beam.

Regarding your 2nd questions: have a look at the link below. It does a good job of illustrating sign conventions on a section cut to the right. The sign convention I've selected is how RISA, STAAD, RAM, and ROBOT all deal with torque. The only outlier I found was SAP2000, which has the opposite sign convention for torque. I still plan to adjust the My and Mz moments eventually.
- https://risa.com/risahelp/risa3d/Content/Common_Topics/MembersBeams%20-%20Results.htm

JWock82:
Another way to think of it is if I apply a positive torque at midspan of a fixed ended beam, I should see a negative torque at the applied to the start of the beam by the supports. This is T1 for the first segment. The fixed end reactions take care of the sign change for us.

JWock82:
Also, I've added a torsion example in the testing and debugging folder for you to review. You can run it to see how the final implementation works.

LcnGnd:
The links you sent me haven't the same positive convention. So, in order to avoid confusion for future developing, I would like to be sure about convention used with torque. 

- Using AddMemberPtLoad, is concentrated torque positive if vector is directed along positive x local axis?
- About torque reactions in the nodes, are they positive if directed along positive X global axis?
- About internal axial, shear, moment and torsion, is positive convention the one indicated in figure below take by https://knowledge.autodesk.com/support/robot-structural-analysis-products/learn-explore/caas/CloudHelp/cloudhelp/2015/ENU/Robot/files/GUID-D1195F2E-6310-4DF6-BE1E-DCCD2A0D121E-htm.html ?

GUID-10F2F6D7-B4BC-4E70-BD56-BD33C1C2EF1A

PR #30, opened 2020-01-05
Update Visualization.py

LcnGnd:
I've introduced a new function called **DeformedShape**  which is able to create 3D visualization of frame's deformed shape. This new function needs two new classes that I've defined: _VisDeformedNode_ and _VisDeformedMember_. 
I'm not yet able to solve a bug. Description by words isn't the best way; I try with images.
Try with a 2D portal frame:

Cattura

You will obtain this deformed shape:

Cattura1

I think that this problem is releated to the function _Deflection_ and also to the code write by me Let's solve this problem and 3D deformed shape's visualization is possible.
Thaks!

### DIFF
--- a/PyNite/Visualization.py
+++ b/PyNite/Visualization.py
@@ -603,6 +603,233 @@ class VisPlate():
     self.lblActor.SetScale(textHeight, textHeight, textHeight)
     self.lblActor.SetPosition((Xi+Xj+Xm+Xn)/4, (Yi+Yj+Ym+Yn)/4, (Zi+Zj+Zm+Zn)/4)
 
+#DEFORMED SHAPE
+class VisDeformedNode():
+  def __init__(self, node, scalefactor, textHeight):
+    
+    #Node position in deformed configuration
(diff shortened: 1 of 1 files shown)


JWock82:
I'll have a look at it when I get a chance. I'm currently working on revising the sign conventions for bending moments, and I'd like to get that change implemented first.

Deflected shapes are difficult to plot. That's because member deflection does not equal member deformation. The deflection is only one component of the overall deformation. You'll notice in your example above that the local-y component of the deformation is correct for every member. It's the other components that are missing missing. The member translations in the local x-direction haven't been accounted for.

LcnGnd:
If I can I will work on 3D deformed shape again, but now I'ver arrived at this point of code.

About sign convention, I think that drawing for internal forces positive convention and for external load positive convention are needed. Only description of the positive convention by means words in the code can create misunderstandings.

Thaks and good work.

JWock82:
I plan to include a sketch in the wiki along with the new sign convention.

PR #40, opened 2020-01-19
Update BeamSegZ.py - Calculation of the axial deformation

LcnGnd:
Calculation of the axial deformation


JWock82:
I already took care of this. I implemented it a little differently. Instead of using 2 separate methods for deformation and deflection, I've calculated the deformation in the deflection method.

PR #41, opened 2020-01-19
Update Member3D.py - Calculation and plotting of the axial deformation

LcnGnd:
Calculation and plotting of the axial deformation


JWock82:
I already took care of this. I implemented it a little differently. Instead of using 2 separate methods for deformation and deflection, I've calculated the deformation in the deflection method.

LcnGnd:
I think there was a misunderstanding because I was speaking about axial deformation (epsilon = N/EA). But if you think that is unuseful, there isn't any problem

PR #42, opened 2020-01-19
Update Member3D.py - Calculation and plotting relative deflections

LcnGnd:
I've introduced the possibility to calculate and plotting relative deflection; it was suggested to me by my professor. It rappresents deflection purified from rigid motion and it can be useful when we have a very big structure with big nodal displacements. In this case, normal deflection diagram could be straight line becaure relative deflections are so small respected to total deformation.
Also, I have seen this function in SAP2000 software too.

### DIFF
--- a/PyNite/Member3D.py
+++ b/PyNite/Member3D.py
@@ -1186,6 +1186,92 @@ class Member3D():
         Member3D.__plt.xlabel('Location')
         Member3D.__plt.title('Member ' + self.Name)
         Member3D.__plt.show()
+    
+#%%   
+    def RelativeDeflection(self, Direction, x):
+        
+        d = self.d()
+        d1y = d[1,0]
+        d2y = d[7,0]
+        d1z = d[2,0]
+        d2z = d[8,0]
+        L = self.L()
+
+        """
+        Returns the relative deflection at a point along the member's length
+        
+        Parameters
+        ----------
+        Direction : string
+            The direction in which to find the relative deflection. Must be one of the following:
+                "dy" = Deflection in the local y-axis
+                "dz" = Deflection in the local x-axis
+        x : number
+            The location at which to find the relative deflection
+        """
+        
+        # Check which axis is of interest
+        if Direction == "dy":
+            
+            # Check which segment "x" falls on
+            for segment in self.SegmentsZ:
+                
+                if round(x,10) >= round(segment.x1,10) and round(x,10) < round(segment.x2,10):
+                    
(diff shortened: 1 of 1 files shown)


PyNite/Member3D.py
```diff
@@ -1186,6 +1186,92 @@ def PlotDeflection(self, Direction):
...
+        
+        Parameters
+        ----------
+        Direction : string
+            The direction in which to find the relative deflection. Must be one of the following:
+                "dy" = Deflection in the local y-axis
+                "dz" = Deflection in the local x-axis
+        x : number
+            The location at which to find the relative deflection
+        """
```
JWock82: The docstring should be directly below the method definition to function properly.

PyNite/Member3D.py
```diff
@@ -1186,6 +1186,92 @@ def PlotDeflection(self, Direction):
...
         Member3D.__plt.show()
+    
+#%%   
+    def RelativeDeflection(self, Direction, x):
+        
+        d = self.d()
+        d1y = d[1,0]
+        d2y = d[7,0]
+        d1z = d[2,0]
+        d2z = d[8,0]
```
JWock82: Use `dyi`, `dyj`, `dzi`, and `djz` for member beginning and ending points. This will keep it consistent with the rest of the code. I've been using `i` and `j` for member ends and `1` and `2` for segments ends.

PyNite/Member3D.py
```diff
@@ -1186,6 +1186,92 @@ def PlotDeflection(self, Direction):
...
+        ax.axhline(0, color='black', lw=1)
+        ax.grid()
+        
+        x = []
+        d_relative = []
+        
+        # Calculate the relative deflection diagram
+        for i in range(20):
+            
+            x.append(self.L() / 19 * i)
```
JWock82:
```suggestion
            x.append(self.L()/19*i)
```
This follows the PEP8 python style guide. Spaces in mathematics should only be between `+` and `-` signs.

PyNite/Member3D.py
```diff
@@ -1186,6 +1186,92 @@ def PlotDeflection(self, Direction):
...
+            d_relative.append(self.RelativeDeflection(Direction, self.L() / 19 * i))
```
JWock82:
```suggestion
            d_relative.append(self.RelativeDeflection(Direction, self.L()/19*i))
```
PEP8

LcnGnd:
Sorry for my errors to writing. Now I think I have solved what you asked me.

PR #47, opened 2020-02-07
Nodal displacements specified by user - settlement of a support

LcnGnd:
I introduced the possibility to apply support's settlements and nodal displacements/rotations at a free node. Because of this new feature, solving method is changed because partioning of matrix K and vectors P and FER is needed.
I hope you like my work and you take it into account for PyNiteFEA develope. 

Thanks!

### DIFF
--- a/PyNite/FEModel3D.py
+++ b/PyNite/FEModel3D.py
@@ -5,7 +5,7 @@ Created on Thu Nov  9 21:11:20 2017
 @author: D. Craig Brinck, SE
 """
 # %%
-from numpy import zeros, delete, insert, matmul, divide, add, subtract, nanmax, seterr
+from numpy import zeros, delete, insert, matmul, divide, add, subtract, nanmax, seterr, shape, asarray
 from numpy.linalg import inv, matrix_rank
 from PyNite.Node3D import Node3D
 from PyNite.Member3D import Member3D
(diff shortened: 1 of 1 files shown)


JWock82:
I've read your code and I understand what it's doing. You've done a great job here! Thank you! Since this is such a major change to the program, I'd like to pull this code into a separate branch, polish it up a little, test it, and then commit it to the master branch. This will take a little more time, but it's important that I thoroughly test such a big change to the program's core solver. I'll set up a new branch in the coming days and pull this in to begin the process.

LcnGnd:
Thanks!!! I'm happy you liked it! 
I'm using PyNite for my master thesis and togheter with my professor we are modeling a girder bridge in order to perform linear elastic analysis and  use them togheter with a monitoring system installed on the bridge. This rappresents the start of a big project. If you are agree with me, I think is correct mention you in my thesis work.

JWock82:
I've merged this code into a new "Pull-Request" branch I created. This will allow me to review and test the code in my own IDE on my computer rather than on GitHub. Once I've finished reviewing and testing I'll merge it into the master branch for everyone to use. Thanks again for your help!

JWock82:
Update: I'm nearly done reviewing your code. I've found a couple issues as I've run test problems on it, but nothing too hard to fix. I think the unknown forces due to known displacements are not being included in the reaction results, but I have yet to verify that. Also, I found the algorithm for partitioning is **very** slow on some of my larger models. I've found a better way to partition by using Python's ability to slice the matrices instead of looping through them. I love Python.

LcnGnd:
I'm using PyNite to model a girder bridge and I viewed the slowness of the code, but my code is a start point for future devolping. About unknown forces due to known displacements they should be included in the reaction results becasue I have tested the code with simply structural models and I have compared PyNiteFEA's results with SAP2000's results. Anyway, good job.

JWock82:
This is now complete and part of the master branch. I still have some more testing I'd like to run, but the tests I've run so far are working great.

JWock82:
> 
> 
> I'm using PyNite to model a girder bridge and I viewed the slowness of the code, but my code is a start point for future devolping. About unknown forces due to known displacements they should be included in the reaction results becasue I have tested the code with simply structural models and I have compared PyNiteFEA's results with SAP2000's results. Anyway, good job.

It should be faster now. Let me know if you see a difference. Also, I verified that the reaction results do in fact include the unknown forces by running a test example. Everything should be working well now.

LcnGnd:
Oh yes, it is faster than before!

PR #57, opened 2020-07-09
Removed duplicate code in RenderLoads function

tamalone1:
The if-else blocks in this function contained nearly identical code in each branch. The two branches were different in only two lines in each of the major for-loops (node loads, point loads, and distributed loads). With so many lines in common, it would be hard to modify the function in the future without mistakes. 

With the addition of a local variable `load_factors` that can be used in either branch, I was able to move the for-loops out of the if-else block completely, and delete the duplicates.

The diff is hard to follow because decreasing the indentation level is shown as a deletion and re-addition. The most helpful comments are at line 165 through 176. 
Please let me know if I didn't explain it clearly.


JWock82:
Thanks for taking care of this. I realized this was redundant when I first coded it, but I was in a hurry to push out the feature. I've been meaning to come back to it and clean it up. I like your use of a single dictionary for both load cases and combinations.

PR #59, opened 2020-07-17
Raise exception when stiffness matrix is singular, with error message

tamalone1:
When the Analyze methods encounter a singular stiffness matrix, it prints an error message and returns None. The program continues and crashes when trying to use the None value. Instead, this change raises an exception instead of returning, stopping analysis immediately and displaying the same error message.

Similarly, when the GetNode and related methods fail to find the element, an exception is raised instead of returning None.

Similarly, when the AddNodeLoads and related methods are passed an incorrect Direction argument, an exception is raised instead of doing nothing.

### DIFF
--- a/PyNite/FEModel3D.py
+++ b/PyNite/FEModel3D.py
@@ -439,6 +439,8 @@ class FEModel3D():
                 
                 # Return the node of interest
                 return node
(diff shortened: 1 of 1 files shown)


JWock82:
Thank you for this great contribution to model validation!

PR #65, opened 2020-08-15
Created test modules for the existing test scripts (12/13)

tamalone1:
I have converted 12 of the 13 test scripts into unittest.TestCases so that they can be run automatically. The test script titled `T Matrix Test` is not done yet. All plots and visualizations are removed.

The "correct" values are taken from the test scripts, if included in the source code, or else from the actual script output. Thus, naturally they all pass (with one exception, see below). These values should be reviewed for correctness.

For floating point results, the acceptable error is set to 0.5%. This value should be set to match the project's goals.

**Note**: Currently, one node of the support settlement test fails, due to a 7.5% difference from the expected value.

### DIFF
--- a/Testing & Debugging/test_2D_frames.py
+++ b/Testing & Debugging/test_2D_frames.py
@@ -23,9 +23,11 @@ class Test_2D_Frame(unittest.TestCase):
         sys.stdout = sys.__stdout__
 
     def test_XY_gravity_load(self):
-        # Create a new model
+        # A First Course in the Finite Element Method, 4th Edition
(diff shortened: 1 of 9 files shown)


JWock82:
The expected results on the support settlement problem came from a textbook that I seem to remember has a lot of rounding errors. I seem to recall the author rounds intermediate results a lot in that book and the rounding error carries over through the whole problem.

JWock82:
I'll review and merge these when I find a moment. Thanks for your help on this effort.

PR #81, opened 2020-12-23
A proof of concept for creating models by reading data files

tamalone1:
I created a few functions for reading data from CSV files to get node, member, support, and load information, then use those to build an FEModel3D. These are a proof of concept for some potential ways to facilitate building models more easily, particularly large ones. 

To demonstrate, I created a script that creates a rectangular grid of nodes, members, supports, and loads and writes them to CSV files. Another script reads those files to build an FEModel3D, analyzes it, and runs visualization. The size of the grid is currently 275 nodes, but is easy to enlarge.

I also made one tweak to FEModel3D.__Renumber, using the enumerate function in place of the manual increment of the id number. This is a clean-up but otherwise not much to note: I don't think there's a speed boost for doing it.


JWock82 (approved):

tamalone1:
If these methods are useful to others, we can move over to a package module at some point. In the meantime, they are just stashed in the example folder.

PR #90, opened 2021-03-22
Element auto-naming and duplicate checks

bjhowie:
Added an option to auto-name all model elements by leaving the 'name' field empty, avoiding the need to manually keep track of element names. This is especially useful for large models.

Also implemented duplicate name checks upon addition of model elements.


JWock82:
Thank you bjhowie. I'll review this code and get it incorporate as soon as I can.

JWock82 (approved):

PR #99, opened 2021-08-14
Check for quads or plates before Rendering Contours.

tamalone1:
I changed the check for `color_map` to check if the model has `Quads` or `Plates` directly. If the model does not have either `Quads` or `Plates` in it, rendering contours will fail with division by zero. This error occurs in `"Simple Beam - Point Load.py"` for example. I think this fix is clearer than checking `color_map` because it hits the root cause of the error. The `color_map` parameter is default=`None` in many places, which makes it easy to accidentally be set to `None`.

I also added a `ValueError` for the `Moment3D.moment` method, for invalid `Direction` arguments.


JWock82 (approved):
Thank tamalone1. I just noticed this issue recently too. This is a much better way to handle it.

PR #103, opened 2021-10-27
Bug fix in setting num_quads for CylinderMesh

tamalone1:
The calculation of `num_quads` using `mesh_size` now includes the radius. 

Before the change, if `mesh_size` > pi, the calculation comes up with `num_quads` = 0 or 1. As the program continues, it eventually fails somewhere else, with a ZeroDivisionError or similar error. Technically, this is due to a unit mismatch of rad/inch instead of inch/inch. The intended calculation is `2*pi*r/m`, not `2*pi/m`.

I think this is the only place this calculation happens this way.

### DIFF
--- a/PyNite/Mesh.py
+++ b/PyNite/Mesh.py
@@ -1066,7 +1066,7 @@ class CylinderMesh(Mesh):
 
         # Determine the number of quads to mesh the circumference into
         if num_quads == None:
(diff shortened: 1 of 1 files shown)


JWock82 (commented):
Thanks for your pull request. Ironically I had just fixed this issue on my local version around the time you submitted this. It's fixed in the latest version v0.0.52.

PR #112, opened 2022-02-06
Tutorials

geosharma:
Thank you for open sourcing your excellent software.

I was trying to learn and created these two additional examples of simple multi-span beams. If you think they would be helpful to others then please add them to the Examples folder.

I am not a structural engineer, but I roughly checked the results with some standard examples for 2 span and 3 span beams examples I found in the internet and it seems like I have modeled the problems correctly.

I also wrote the a separate plot and save function because floating windows are a bother in my i3 window manager, as I am not really good with the configurations. I guess there is no way to save the vtk window without displaying.

I should have named by branch differently but contributing to examples only hit me after the fact.

Thank you once again for your time and effort and releasing this as open source. Like I mentioned I am not a structural engineer and have only dabbled in ready-made FEA programs like Opensees (from my own interest) and RS2 (work related), but if there is any way to help with small things please do let me know.


JWock82:
Sorry for the delay in getting back to you. I'll have a look through this. Thank you.

scr/example-01b_simple-beam-3-span_point-load.py
```diff
@@ -0,0 +1,303 @@
...
+from PyNite import Visualization
+from pathlib import Path
+import matplotlib.pyplot as plt
+from itertools import accumulate
+
+# Updated parameters for this plot
+plt.style.use("../plt/figstylefile.mplstyle")
+
+
+class Section:
```
JWock82: I've been planning to integrate a "Section" class directly into PyNite for some time. I don't like putting it in the examples as many users may not be very good at object oriented programming. I'd like to keep the examples simple.

scr/example-01b_simple-beam-3-span_point-load.py
```diff
@@ -0,0 +1,303 @@
...
+    def __init__(
```
JWock82: This line is formatted different from the rest of PyNite. That's fine, but I'd rather pull it into a separate branch than "master" so that I can format it to match before pulling it into the master branch.

scr/example-01b_simple-beam-3-span_point-load.py
```diff
@@ -0,0 +1,303 @@
...
+    path_root = Path(r"../")
+    # path_data = path_root.joinpath("data/")
+    path_plt = path_root.joinpath("plt/")
+    path_report = path_root.joinpath("rpt/")
+    fname_vtk = "example-01b-vtk.png"
+    fname_fig = "example-01b.png"
+    fname_rpt = "example-01b_rpt.pdf"
+    ffigvtk = path_plt.joinpath(fname_vtk)
+    frpt = path_report.joinpath(fname_rpt)
+    ffig = path_plt.joinpath(fname_fig)
```
JWock82: I don't like this kind of stuff in the examples. This has to do with preferences for how/where to save the files and very little to do with using PyNite and will turn off a lot of users i they see it.

scr/example-01b_simple-beam-3-span_point-load.py
```diff
@@ -0,0 +1,303 @@
...
+    # beam FEM model
+    sbeam = FEModel3D()
+
+    # define section: section_name, cross-section area, Iz, Iy, J, E, G
+    # J=None, E=29000 ksi, G=11400 ksi are optional parameters
+    section_1 = Section("section-1", 20, 150, 100)
+
+    # Point load [kip]
+    pt_load = -10
+    # lengths of each span [ft]
```
JWock82: This could be a lot simpler now that physical members are supported in PyNite.

JWock82 (commented):
Thank you for this pull-request, but overall I think these examples are a little too complex for most structural engineers. I'd like to keep the example folder very basic to help new users get up and running with minimal coding effort.

PR #116, opened 2022-02-23
Updated the outdated method names and keyword parameters in the Examples scripts, and in Member3D.

tamalone1:
In the example scripts, it was references to `RenderModel` and the keyword change to `annotation_size`. In `Member3D`, there was an internal call to the deprecated `Shear` method.


JWock82 (approved):

JWock82:
Thanks!

PR #125, opened 2022-03-30
Repair method

### LINKED ISSUE #100: Automatic subdivision of members with intermediate nodes
I'd love to see functionality where the solver automatically recognizes the presence of an intermediate node within the length of a member, and internally divides the member into smaller sub-members. With that functionality, an intermediate framing member will behave correctly, without having to subdivide the member by hand. This functionality is really nice to have in commercial solvers.

maderero:
New feature Fixes #100

### DIFF
--- a/PyNite/Member3D.py
+++ b/PyNite/Member3D.py
@@ -1,6 +1,6 @@
 # %%
-from numpy import array, zeros, matrix, add, subtract, matmul, insert, cross, divide
-from numpy.linalg import inv
+from numpy import array, zeros, matrix, add, subtract, matmul, insert, cross, divide, multiply
+from numpy.linalg import inv, det, norm
 from math import isclose
 from PyNite.BeamSegZ import BeamSegZ
 from PyNite.BeamSegY import BeamSegY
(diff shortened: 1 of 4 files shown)


maderero:
🎉🎉🎉

PyNite/Member3D.py
```diff
@@ -1764,4 +1764,83 @@ def _segment_member(self, combo_name='Combo 1'):
...
+    def coordinates(self):
+        return (node.coordinates for node in self.nodes)
+
+    def intersection(self, other, tolerance=1e-3, virtual=False):
+        if virtual:
+            return self.intersection_virtual(other, tolerance)
+
+        return self.intersection_real(other, tolerance)
+
+    def intersection_virtual(self, other, tolerance=1e-3):
```
JWock82, line 1782: I'm trying to comment this code to bring it into the main/master branch of PyNite. I'm not sure what this method does. The math in here is pretty abstract to me. Would you mind providing some comments to help guide me through what's taking place?

maderero: Yes I will be back shortly with an explanation.

maderero: This is a port of the `intersection_line_line_3d` function from [mozman's ezdxf](https://github.com/mozman/ezdxf/blob/3a4be02d937dc5744785763dd7a7d2a8699eef19/src/ezdxf/math/_construct.py#L146). There is a reference to the math involved [here](https://www.realtimerendering.com/intersections.html#I304).

The math checks for intersections in two rays (virtual intersections). If the rays do not intersect, we can assume that the line segments that make up those rays will not intersect either. If there is a virtual intersection, then we determine if there is a pair of the line segments that exist at the point of the virtual intersection. If not, the line segments have the potential to intersect, but ultimately do not. If there is, we have found a real intersection.

maderero: Virtual intersections denote intersections that occur between two rays.
Real intersections denote intersection that occur between two line segments.

maderero:
See [this commit](https://github.com/maderero/PyNite/commit/2e9dcaaba575850a1711415bc4e26466c81a1cab) for more info. I have not done much testing on this since I am a bit busy today.

maderero:
There is a bug in the `virtual_intersection` method for members that are perpendicular. 
The tolerance can not be set at or above 1 otherwise no intersections will be detected. 
This is due to the calculation that is done when calculating the `denominator` value.
i.e. `A x B = |A||B| sin θ` for `θ=pi/2` and normalized vectors `A` and `B` gives `A x B = 1`.
I will propose a change later today.

PR #126, opened 2022-03-31
merge_duplicate_nodes speed-up

maderero:
Current method iterates through all elements (members/springs/plates/quads) on every pass comparing nodes. 

In this pr, node usage is checked ahead of time to reduce runtime.
Tested with 50000 nodes and 200 elements and noted a substantially reduction in runtime (~45 sec to ~8 sec on my machine).
Also added a new method to the Node3D class (`Node3d.distance`) to calculate the distance to a secondary node.

### DIFF
--- a/PyNite/FEModel3D.py
+++ b/PyNite/FEModel3D.py
@@ -471,7 +471,8 @@ class FEModel3D():
         for element_attr in element_attrs:
             for element in getattr(self, element_attr).values():
                 for node_attr in node_attrs:
-                    if (node := getattr(element, node_attr, None)):
+                    node = getattr(element, node_attr, None)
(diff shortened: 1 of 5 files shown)


maderero:
Looks like checks are failing due to walrus operators. I'll rewrite those bits after lunch.

JWock82:
From the checks above it looks like some of the syntax is not compatible with python 3.6 and 3.7. I'm trying to keep PyNite compatible with python 3.6+. Could you expand the details, take a look, and revise your pull request?

maderero:
Alright I'll keep that in mind in the future. Fixed the issue!

maderero:
Deleted branch on accident

PR #127, opened 2022-04-12
Speedup repair method

maderero:
1) Fixes [this](https://github.com/JWock82/PyNite/pull/125#issuecomment-1096798356).
I split up the `tolerance` kwarg in the `repair` method and set a maximum value for the tolerance in the `intersection_virtual` method to fix this issue.
2) Speeds up repair method by skipping duplicate vector calculations (these are expensive). 
Tested using 25-100 horizontal and 25-100 vertical members with good results; I was seeing 20-38% reduction in runtime.
3) Added some annotation in the `repair` method so it is easier to follow the logic. 
4) Change some naming.


maderero:
closed on accident; reopened

Issue #144, opened 2023-01-28
Add methods to Member for returning result arrays


connorferster:
**Is your feature request related to a problem? Please describe.**
In order to generate a factored load envelope, I need access to the results arrays of various load cases/combinations. I can generate these on my own but thought that it would be a welcome addition and possibly a stepping stone for future features in #101.

I am happy to create the PR for this!

**Describe the solution you'd like**
The suggestion is to create the following methods onto `Member`:

* `.axial_array`
* `.torque_array`
* `.shear_array`
* `.moment_array`
* `.deflection_array`

They could be implemented similar to the hard-coded approach you have in `.plot_shear`, etc.

```python
def shear_array(dir: str, n: int, combo_name: str) -> np.ndarray:
    """
    Returns an array describing the shear result...
    ...
    """
    L = self.L
    x_arr = np.linspace(0, L, n)
    y_arr = np.array(
        [self.shear(dir, combo_name) for x in x_arr]
    )
    return np.array([x_arr, y_arr])
```

These methods could then replace those sections of code in the `.plot_...` methods and would allow for the `.plot_...` methods to have a user-specified `n` indicating the number of points they would like to discretize the plot with, which would be good because sometimes 20 points is just not enough and the plots end up looking distorted and confusing.

**Describe alternatives you've considered**
Generating the arrays myself

**Additional context**
Having access to results arrays allows for plots to be generated in any plotting software, not just matplotlib. By generating the resulting arrays conveniently, it allows for a little bit extra extensibility both for plotting and post-processing.

JWock82:
I like this idea. I think there's a way to do it with minimal extra lines of code. What if we had a boolean argument in the `plot_shear` method named 'to_array' (defaults to `False`) that changed the output of the method to an array instead of a matplotlib plot? I agree that 20 points is often just not enough too. I've been meaning to add an 'n' argument to `plot_shear` as well.

JWock82:
It seems like all it would take is an `if` statement placed conveniently to ignore the plotting commands and return the array directly.

PR #145, opened 2023-01-31
Results arrays (#144)

connorferster:
Addresses #144 with the following changes:

* Added an "array" method for `Member3D` for the following:
  * `shear_array`
  * `moment_array`
  * `torque_array`
  * `axial_array`
  * `deflection_array`
  * `rel_deflection_array`
* Added an `n_points` argument to the "plot" methods to allow for varying the resolution of the plot (added the argument as the final keyword argument to not break any existing code).
*  Added an example for the use case of plotting factored envelopes of varying number of points (Examples/"Simple Beam - Factored Envelope.py")
* Fixed bug in `Member3D.rel_deflection` where the actual `LoadCombo` object was being passed to `Member3D.d` instead of the combo name.
* Fixed various typos in doc strings of "plot" methods

### DIFF
--- a/Examples/Simple Beam - Factored Envelope.py
+++ b/Examples/Simple Beam - Factored Envelope.py
@@ -1,0 +1,54 @@
+# Units used for the model in this example are inches and kips
+
+# Import `FEModel3D` from `PyNite`
+from PyNite import FEModel3D
+from matplotlib import pyplot as plt
+import numpy as np
+
+# Import 'Visualization' for rendering the model
+from PyNite import Visualization
+
+# Create a new finite element model
+SimpleBeam = FEModel3D()
+
+# Add nodes (14 ft apart)
+SimpleBeam.add_node('N1', 0, 0, 0)
+SimpleBeam.add_node('N2', 14*12, 0, 0)
+
+# Add a beam with the following properties:
+# E = 29000 ksi, G = 11400 ksi, Iy = 100 in^4, Iz = 150 in^4, J = 250 in^4, A = 20 in^2
+SimpleBeam.add_member('M1', 'N1', 'N2', 29000, 11400, 100, 150, 250, 20)
+
+# Provide simple supports
+SimpleBeam.def_support('N1', True, True, True, True, False, False)  # Constrained for torsion at 'N1'
+SimpleBeam.def_support('N2', True, True, True, False, False, False) # Not constrained for torsion at 'N2'
+
+# Add a downward point load of 5 kips at the midspan of the beam
+SimpleBeam.add_member_pt_load('M1', 'Fy', 5, 4*12, 'D') # 5 kips Dead load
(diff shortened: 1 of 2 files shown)


connorferster:
@JWock82 

Let me know your thoughts on this and if you would like any other additions to the code base to support this proposed change.

JWock82:
I like this. It seems like `shear_array` and `plot_shear` are similar to each other and this could be done in `plot_shear` with an argument that switches the output from `matplotlib` to an array. If you want to merge the methods together feel free and then issue another PR. Otherwise I'll merge this PR and then merge the methods myself when I find a spare moment in the next week or so.

connorferster:
Glad you like it!

Agreed, the "array" functions and the "plot" functions are similar and the implementation code is repetitive. I hear what you are saying on adding a boolean flag to the plot functions. My only misgiving with such an approach is that it would make the return type of the plot functions inconsistent which I think is unexpected. If I was cruising the API to see what method to use to get a result array, the plot methods would not be my first guess.

However, I think that reducing the amount of repetition would be good because it reduces the quantity of code requiring maintenance. Let me know what you think of this counter proposal:

Add a method like this:

```python
def _results_array(self, action: str, n_points: int, direction: Optional[str] = None, combo_name="Combo 1") -> np.ndarray:
    """
    Returns the results array according to the parameters.

    action: {"shear", "moment", "torque", "axial", "deflection", "rel_deflection"}
    n_points: the number of points
    direction: the direction of action to retrieve
    combo_name: the load combination to retrieve
    """
    action_function = getattr(self, action)
    L = self.L()
    x_arr = linspace(0, L, n_points)
    if direction is None:
            y_arr = array(
                [action_function(x, combo_name) for x in x_arr]
            )
    else:
            y_arr = array(
                [action_function(direction, x, combo_name) for x in x_arr]
            )
    return array([x_arr, y_arr])
```

Then, the implementation of each array function simply becomes:

```python
def shear_array(direction: str, n_points: int, combo_name="Combo1") -> array:
    """
    ...doc string
    """
    return self._results_array("shear", direction, n_points, combo_name)
```

The plot functions could then call the `_results_array` function in a similar fashion. This would leave, effectively, only one method that requires maintenance. If you wanted to take it further, you could do the same with the plot methods: abstract it into a single `_results_plot` method (for example) and then each plot method would simply call that, each with the appropriate inputs. Then you would only have two methods that need maintaining and you have created convenient and intuitive API endpoints for the user (i.e. `plot_shear` and `shear_array`, etc.) Since these methods would be "private" methods, and only called internally by other methods, you could also do away with a lot of error checking (like checking if the `action` string passed was a valid `action` name).

Let me know your thoughts on this and I can put it together.

JWock82:
I see what you're getting at, and I'll probably move it that direction. I've merged your PR for now.

Issue #146, opened 2023-02-09
PendingDeprecationWarning: matrix subclass to be deprecated


connorferster:
**Describe the bug**
When running an analysis, I was recently greeted with the following warning:

```python
PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
    dirCos = matrix([x, y, z])
```

**To Reproduce**
I think this comes up with any analysis.

**Proposed solution**
I think this can be fixed just by substituting `array` for `matrix` to build a regular `numpy.ndarray`. I will submit a PR for this fix.

connorferster:
Resolved in #147

PR #147, opened 2023-02-10
Replace numpy.matrix with numpy.array

connorferster:
In answer to #146, this PR replaces a handful of instances of `numpy.matrix` with `numpy.array` where they appear in Member3D.py and Plate3D.py. Also changed one `*` operator to `@` now that matrix multiplication is no longer assumed.


JWock82:
Great! I've been slowly cleaning those instances up as I go. A few years ago numpy said they were deprecating "Matrix" and I just haven't gotten around to cleaning up all the instances - lazy I know.

PR #166, opened 2023-07-19
Allow Selective Analysis Of Load Combos

SoundsSerious:
I am using a significant set of load combinations for aerodynamic analysis (permutations of speed,rotation,angle) and I wanted a way to choose which to run running all of them takes about 15minutes.

I have added a `combos:list = None` keyword to each `analyze` function along wih appropriate documentation.

The analysis function now checks directly after the LoadCombos iteration if combos is specified the combo is in combos the case will proceed, otherwise continue is used to bypass unspecified combos

springs and member are activted for the combo in question or all if `combos` is None.

combos is also passed to `_check_static` and `_calc_reactions` so that only the run combos are checked.

I have run the testing suite on this and it appears fine. I haven't added any new tests.

### DIFF
--- a/PyNite/FEModel3D.py
+++ b/PyNite/FEModel3D.py
@@ -1968,11 +1968,12 @@ class FEModel3D():
         D2 = atleast_2d(D2).T
 
         # Step through each load combination
-        for combo in self.LoadCombos.values():
-            
+        for combo_name,combo in self.LoadCombos.items():
+
             #Check if specific combos and if this combo is selected
-            if combos is not None and combo not in combos:
-                continue
+            if combos is not None and combo_name not in combos:
+                continue     
+
 
             if log:
                 print('')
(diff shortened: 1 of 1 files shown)


SoundsSerious:
Ok now we're looking good. I was comparing the combo instance to the key so it wasnt evaluating.

I reran the test cases and they are continuing too work but considering the error I just made it might be a good idea to cover this case.

SoundsSerious:
Sure thing, I think that should be good to go.

JWock82:
I merged this, and then reverted it, realizing that it had some conflicts with other changes I am making that I hadn't committed yet.

I did read through your code and I like the utility of this feature. I'm going to rework it a little to work with load combination tags. That way you'll be able to apply a tag to your load combination and then run the analysis for the tags you specify.

SoundsSerious:
sounds good, It definitely makes sense to take time and do it right.

PR #169, opened 2023-08-04
Add fix for 3D cylinder mesh:

SoundsSerious:
>allow origin reading for input on coordinate
>do not adjust number of node spacing in loop as that can result in unpredictable behavior
>ensure at least one mesh element will be made


SoundsSerious:
I ran into an issue where mesh elements weren't being added when the mesh size was larger than the height, so I added some protection to ensure at least one ring of mesh is made. I also noticed an issue where the discretization change dynamically as the cylinder mesh was progressively made as `y` changed so I moved this before the loop.

I see there are some "conflicts" but our code is exactly the same so I've removed my duplication for selection of `y-origin` per axis selection.

PyNite/Mesh.py
```diff
@@ -1273,7 +1273,14 @@ def generate(self):
...
         h = self.h
-        y = self.origin[1]
+
+        if self.axis == 'Y':
+            y = self.origin[1]
+        elif self.axis == 'X':
+            y = self.origin[0]
+        elif self.axis== 'Z':
+            y = self.origin[2]        
+
```
JWock82: I agree with these changes, but my copy of Pynite already has them. Someone else recently pointed this out and I fixed it. I need to update the PyPI version to include this change.

PyNite/Mesh.py
```diff
@@ -1283,12 +1290,17 @@ def generate(self):
...
+        
+        #Element height in the vertical direction
+        h_y = height/n_vert    
+
         # Mesh the cylinder from the bottom toward the top
         while round(y, 10) < round(h, 10):
-            
-            height = h - y                  # Remaining height to be meshed
-            n_vert = int(height/mesh_size)  # Number of times the plate height fits in the remaining unmeshed height
-            h_y = height/n_vert             # Element height in the vertical direction
```
JWock82, line 1304: I agree with adding the "max" function to ensure that at least one course of plates is being generated.

As long as `h_y` is constant on each iteration, it makes sense to move these variables assignments outside the loop. However, I think it's best to leave the definition of `height`, `n_vert`, and `h_y` in the loop. At some point, I may add control points to the cylinder mesh, allowing `h_y` to vary as the cylinder is generated. This could be useful for hydrostatic loads or changes in shell thickness that cause a discontinuity partway up the cylinder. Right now we're forced to generate two separate cylinders.

JWock82 (changes requested):
I recommend leaving most of the code as is, except let's go ahead and add the "max" function that guarantees at least one course of plates is generated. See my comments for detailed explanations.

JWock82 (approved):

PR #171, opened 2023-08-09
Quad Stress Clarification & Input Checks

SoundsSerious:
I have a need to predict failure on mesh elements (which I am fairly rusty on) and have taken a stab at improving the documentation the `Quad.member()` function. I thought it would be more useful to put this in a PR vs an Issue so as to reduce your workload.

I've studied your code as well as some background on MITC4, and believe your `member()` function is designed to return the stresses / shear which I need for my use case. 

I've made two changes here which shouldn't affect results

1. Add Bound Protection to shear,moment,member functions to ensure (r,s) are within -1,1

2. Stress Description to Quad3D.member() function as to what I think is appropriate (but I may be mistaken)

Additionally:
I noticed as well some of my element mesh spacing is off but right now I suspect my own code but I can follow up here so as to consolodate changes. I think this may have to do with some of the origin issues previously solved.


SoundsSerious:
Hmm I may have also included some cylinder mesh changes that we've already discussed. I think I have deleted the file vsv the diff so I'm closing this and will resubmit this PR without those change to ensure that Mesh.py doesn't go missing

PR #194, opened 2024-05-05
Update parameter names in FEModel3D methods for consistency of API

connorferster:
Hi @JWock82,

I noticed that there were some breaking changes to the API with recent versions but that the updates were not applied to all of the methods in `FEModel3D`. I have updated the rest of the parameter names across the `FEModel3D` object by following what seemed to be the desired new convention. Here is how I understand it:

* When creating a new object (e.g. node, member, section, etc.), the method that creates it (e.g. `.add_node`) will accept a parameter called `name` to set the name of the object in the model.
* When creating an assignment to an existing object (e.g. a load to a member), then the parameter will be named something like `member_name` (as opposed to `Member`) to make it clear that this parameter is to accept a name and not a `Member` object itself.

I ran the test suite and all tests pass. However, I notice that the test suite does not use any keyword arguments when the methods are called. The tests passing will reflect that the parameter name changes I made in the function signatures have been updated in the function implementations and not that I have followed the naming convention perfectly (although, from manual review, I think I have).

Let me know your thoughts on this?

### DIFF
--- a/PyNite/FEModel3D.py
+++ b/PyNite/FEModel3D.py
@@ -982,9 +982,9 @@ class FEModel3D():
         :raises ValueError: _description_
         """
             
-        # Validate the value of Direction
+        # Validate the value of direction
         if direction not in ('DX', 'DY', 'DZ', 'RX', 'RY', 'RZ'):
-            raise ValueError(f"Direction must be 'DX', 'DY', 'DZ', 'RX', 'RY', or 'RZ'. {direction} was given.")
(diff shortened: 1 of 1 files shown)


connorferster:
Here is an example (from the docs) of what I meant when I referred to changes to the API not being propagated across the `FEModel3D` methods:

image

connorferster:
I also change the `Direction` parameter to `direction`.

PR #195, opened 2024-05-06
Refinement and Flexibility Enhancements for Tension/Compression Convergence Checks in Structural Analysis

Jeroen124:
### Overview
This pull request introduces several key improvements to the structural analysis logic focusing on the tension/compression convergence checks. These changes aim to increase the flexibility of our simulation parameters and reduce code duplication, resulting in cleaner and more maintainable code.

### Changes Made
1. **Refactored Convergence Checks**:
   - Implemented a loop to handle all directional checks (`DX`, `DY`, `DZ`, `RX`, `RY`, `RZ`) in a unified manner using dynamic attribute access. This approach eliminates repetitive code blocks and simplifies modifications.
   - Implemented logic to reassess and potentially reactivate members based on new conditions, ensuring that members previously deactivated can become active again if the situation changes.
   
2. **Introduced Tolerance Parameters**:
   - Added `spring_tolerance` and `member_tolerance` as optional parameters to control the activation and deactivation of springs and members based on specific thresholds.
   - Integrated these tolerances into the convergence checking logic, allowing for more realistic and stable simulation outcomes.

3. **Updated Main Analysis Function**:
   - Enhanced the `analyze` function to accept and propagate the new tolerance parameters to the convergence checking subroutine.

### Benefits
- **Increased Flexibility**: Users can now specify tolerance values to prevent minor fluctuations from affecting the analysis, leading to more stable and realistic results.
- **Improved Code Quality**: The refactoring reduces complexity and enhances the readability and maintainability of our codebase.

### DIFF
--- a/PyNite/Analysis.py
+++ b/PyNite/Analysis.py
@@ -3,13 +3,14 @@ from PyNite.LoadCombo import LoadCombo
 from numpy import array, atleast_2d, zeros, subtract, matmul, divide, seterr, nanmax
 from numpy.linalg import solve
 
+
 def _prepare_model(model):
     """Prepares a model for analysis by ensuring at least one load combination is defined, generating all meshes that have not already been generated, activating all non-linear members, and internally numbering all nodes and elements.
 
     :param model: The model being prepared for analysis.
     :type model: FEModel3D
     """
-    
+
     # Reset any nodal displacements
     model._D = {}
     for node in model.Nodes.values():
@@ -23,8 +24,9 @@ def _prepare_model(model):
     # Ensure there is at least 1 load combination to solve if the user didn't define any
     if model.LoadCombos == {}:
         # Create and add a default load combination to the dictionary of load combinations
-        model.LoadCombos['Combo 1'] = LoadCombo('Combo 1', factors={'Case 1':1.0})
-    
+        model.LoadCombos['Combo 1'] = LoadCombo(
+            'Combo 1', factors={'Case 1': 1.0})
+
     # Generate all meshes
     for mesh in model.Meshes.values():
         if mesh.is_generated == False:
@@ -34,14 +36,15 @@ def _prepare_model(model):
     for spring in model.Springs.values():
         for combo_name in model.LoadCombos.keys():
             spring.active[combo_name] = True
-    
+
     # Activate all physical members for all load combinations
     for phys_member in model.Members.values():
         for combo_name in model.LoadCombos.keys():
             phys_member.active[combo_name] = True
-    
+
     # Assign an internal ID to all nodes and elements in the model. This number is different from the name used by the user to identify nodes and elements.
     _renumber(model)
+
 
 def _identify_combos(model, combo_tags=None):
     """Returns a list of load combinations that are to be run based on tags given by the user.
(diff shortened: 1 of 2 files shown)


JWock82:
I really like the simplification of using attributes for the degrees of freedom. Very well done. Can I ask why you're using a spring tolerance? Is this to deal with computational precision errors triggering springs to be active?

I'd prefer not to reformat so much of the code in `Analysis.py`. Most of Pynite uses PEP8's first example (aligned with open delimiter) shown here: [Indentation](https://peps.python.org/pep-0008/).

connorferster:
The reformat has some quirks that I find strange. Normally a reformat would be done on the entire code base using a pre-determined code-formatting tool (e.g. Black or Ruff). @Jeroen124 Did you do this reformat manually?

@JWock82 What are your thoughts on using a code-formatting tool to avoid needing future decisions on where line breaks are made?

Jeroen124:
@connorferster  @JWock82  Sorry for the formatting issues. My VS-code had some settings off. I will adjust Analysis.py without the reformatting. 

@JWock82 I am using Pynite as a quick-simple hand calculations crosscheck, the problem i am running into is in pushover analysis, due to shortering of primary steel frame, some bracing combinations can both become in compression. And therefore are removed after the first linear step, this will become a problem, because then no more bracing is longer applied. Another solutions instead of a tolerance can be, pretension or incremental loadsteps. 

I will get back to the formatting issue tomorrow and update analysis.py

JWock82:
I see what you're doing. There are cases when a structure becomes unstable due to members turning off in a tension-only analysis. Like an inverted V-braced frame with only gravity loads. I've run into this issue in commercial software before. I know one popular program uses "Euler-buckling" members to deal with this issue. It basically sets the tolerance as the member's Euler buckling load. Those members can hold tiny amounts of compression before they switch off. This prevents instability from developing under negligible compression loads in tension-only members.

I think a more theoretically correct solution is to apply the load incrementally in small steps. This allows the program to slowly ascend the structure's elastic load-deflection curve until we hit the nonlinear portion. Using a small load step helps us identify the correct "exit ramp" from linear to nonlinear behavior. Using this approach would allow members to turn back on as the structure starts to deform and destabilize, engaging any braces that experience tension on the subsequent load step. The P-&Delta; analysis takes this approach. Try running a P-&Delta; analysis.

JWock82:
> The reformat has some quirks that I find strange. Normally a reformat would be done on the entire code base using a pre-determined code-formatting tool (e.g. Black or Ruff). @Jeroen124 Did you do this reformat manually?
> 
> @JWock82 What are your thoughts on using a code-formatting tool to avoid needing future decisions on where line breaks are made?

I think a code-formatting tools would be a good idea. I'm not familiar with them myself.

Jeroen124:
@JWock82 & @connorferster maybe we should first implement Ruff for example for formatting? So that we can have uniform formatting. Otherwise i have to reformat the analysis.py by hand. I think when we have standard settings, everyone can follow these formatting rules, it will improve the readability.

Jeroen124:
@JWock82 It was a long time ago i made this PR, i have mannually reformatted back the code to the old style. I think it can be merged now. :)

JWock82:
> @JWock82 It was a long time ago that I made this PR. I have manually reformatted the code back to the old style. I think it can be merged now. :)

Thanks @Jeroen124. I will have a look at this as soon as I can. I'm currently working on some code for an improved plate element, and I want to merge that code into the main branch before I merge this code to make sure it's all compatible. I have made a few changes to the `Analysis.py` file, which may affect this PR. I appreciate your contribution.

PR #196, opened 2024-05-07
Added check and conversion for non-numerical data coming in to sig_fig_round

connorferster:
I tried running the new version to see the new PyVista display. It took some work. Once I got the installs figured out, I found that there seemed to be a bug in the function `sig_fig_round`. It was being passed a `text_label` and it was trying to perform math functions on the value. It seemed that a conversion step was necessary to convert the `text_label` to a `float`. 

I added the conversion into `sig_fig_round` and then the display worked. And...it looks nice! I love that it is not spawning a separate process that hangs your kernel like VTK does. Thank you!

### DIFF
--- a/PyNite/Rendering.py
+++ b/PyNite/Rendering.py
@@ -1250,6 +1250,13 @@ def _PrepContour(model, stress_type='Mx', combo_name='Combo 1'):
                 node.contour = sum(node.contour)/len(node.contour)
 
 def sig_fig_round(number, sig_figs):
+    # Check for strings or other convertible data types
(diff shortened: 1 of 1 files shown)


JWock82:
Thanks Connor. As `Pyvista` is a new feature, there may still be a few bugs in there I have not caught yet. Keep an eye out for them. It's been a major improvement to the rendering code. Much simpler to implement than directly using `VTK` which required me to manage the whole rendering pipeline. `Pyvista` manages all that for you. This will make future development of the rendering code much simpler.

Issue #192, opened 2024-04-22
Quad results in global coordinate system 


bjhowie:
As I understand it, at the moment you can only extract results from quads (and presumably plates) in the quad's local coordinate system. 

I assume it's a trivial process to convert results into the global coordinate system using the quad's transformation matrix, but I'm a little unfamiliar with the mathematics.

If it's something that's straightforward to implement, it'd be great to see it in a future release.

bjhowie:
Or as another thought, each quad could have a local axis vector defined (or use an auxiliary node) that gets projected onto the plane of the quad to form the quad's local X axis instead of just using the node3->node4 vector as is currently done. I assume the 3-4 vector is just an arbitrary selection anyway?

I'm happy to have a go at implementing if you think it could work...

JWock82:
I think it’s pretty simple. I believe you just take the quad results in each direction and multiply them by the appropriate direction cosines for the quad. The direction cosines are the first 3 rows (dircosX, dircosY, dircosZ) and first 3 columns of the transformation matrix, T, which PyNite already can calculate for you. I’d be happy to have your help.

JWock82:
I've decided to take a stab at this. It's a feature I've been meaning to get to. I'll let you know what I come up with.

bjhowie:
Thanks, it looks like my attempt failed. It seems to work OK for rectangular elements which I had originally tested it with, but results diverge pretty quickly with skewed quads.

Below are two deflection plots, both using the custom local axes method from my pull request. The rectangular(ish) quad plot looks quite good, the other with skewed quads is incorrect. Both deflection plots look OK with the previous local axis system, but obviously the quad moment and shear plots are not correct for the skewed quad example.

![Screenshot 2024-05-08 204510-rect](https://github.com/JWock82/PyNite/assets/45848617/c3907d63-b40c-433e-98d9-2ed1e8ae94a6)

![Screenshot 2024-05-08 205637-skewed](https://github.com/JWock82/PyNite/assets/45848617/7504a15d-4715-4a70-9892-69603be94c23)

JWock82:
I've created a new branch `global_plate_stress` for development of this feature. Feel free to check out and test out that branch and see if it's behaving as you expect. It seems to be working, but I'd like to double check my "theory" behind the stress transformation before publishing it to the main branch. I have only done it for shears (QX and QY) and moments (MX and MY), not displacements yet.

bjhowie:
[...]

PR #197, opened 2024-05-07
Custom Quad Local Axes & Improvements to VTK Contour Rendering

bjhowie:
I'm aware that the VTK rendering in the Visualization module is being deprecated, but I've made some changes for my own project that I think dramatically improve the look of the quad and plate contour plots and would be good to include in PyNite.
Namely, I have added a subdivision layer that improves the resolution of the underlying scalar field, resulting in smoother looking plots. I've also added an option to use banded contours instead of the default continous colour scale. By default, I've also added an edge extraction filter in order to better see the outline of the individual plates.

Following the issue raised previously (#192), I've also implemented user-defined quad local axes via the use of a vector defined for each quad that gets projected onto the quad to form the local x axis. The local y & z axes are defined as before.
There is however a situation where the defined vector is normal to the plane of the quad, in which case another vector needs to be chosen to form the local x axis. I have just defaulted to using the previous definition (the node3 -> node4 vector) in such cases, and also when the local axis vector is set to 'None'. My preference would be to try and align the local y axs with the global y axs when it isn't possible to align the x axes, but this is less backward compatible. By default, the quads now align their local x axis to the global x axis, requiring a slight change to the tank test case telling the quad to use the previous local axis definition.

I have also stored the materal name as an instance attribute for quads and plates to make it consistent with the Member3D class.

All tests are passing.


bjhowie:
Actually, hold fire on this one. I'm getting some odd deflection results on one of my models, but having trouble reproducing.

bjhowie:
Issue seemed to be that I didn't update the Quad T matrix to use the new local axis orientation. Looks to be working much better now.

Below is a comparison of a rectangular vs skewed quad raft footing example. There is a bit of a difference in the peak sagging moment at the point load locations, but I'm thinking this is due to the different meshing, not due to the axis transformation. Away from the point loads, the shape of the plots look pretty similar. Deflections all look good now too.

raftplot_comparison

JWock82:
> Issue seemed to be that I didn't update the Quad T matrix to use the new local axis orientation. Looks to be working much better now.
> 
> Below is a comparison of a rectangular vs skewed quad raft footing example. There is a bit of a difference in the peak sagging moment at the point load locations, but I'm thinking this is due to the different meshing, not due to the axis transformation. Away from the point loads, the shape of the plots look pretty similar. Deflections all look good now too.
> 
> raftplot_comparison

I'm not sure why you're comparing local Mx moments to global MX moments for a skewed mesh. Those should not be the same.

connorferster:
@bjhowie and @JWock82 

I have been thinking about the visualizations. The VTK is such a great visualization and I think it would be unfortunate to effectively discard all of that effort by deprecating it. My suggestion is to instead to move the visualization engines to separate packages then folks can use the engine most suited to their use case (e.g. VTK is probably best for creating desktop apps).

SoundsSerious:
@connorferster I dont nessicarily think these changes will be lost by a switch to PyVista since there is inter-operability.

https://tutorial.pyvista.org/tutorial/06_vtk/index.html

I personally haven't delved into VTK development since the API seems somewhat complicated, but pyvistas interface seems easy enough. I think as a strategy it could be useful for lowering the bar for contributions from those who aren't VTK proficient but can handle numpy / matplotlib style functions.

bjhowie:
I'm going to close this PR as the quad local axis changes referenced the old MITC4 elements.

I think the contour rendering changes are valuable, but it's unclear whether incorporating new features using VTK is the direction the project wants to head in. I had played around with PyVista but found it to be quite inflexible compared to raw vtk, even though the vtk API is awful. It would be nice if we could find a way incorporate vtk features on top of a PyVista base.

I can resubmit the contour rendering code in future if need be.

PR #206, opened 2024-07-25
Mesh Modifications & Load Combinations

SoundsSerious:
- added an additional selection of `load_combo` in addition to `combo_tags` as a union of input choices
- added an angular offset to cylindrical / annular meshes based on the start node
- clean up merge_duplicate_nodes to fit 80 character wide standard by selectiong n1,n2 prior to replacement logic (nodes removed after identification so it should be the same)
- added a KDTree duplicate node identification method
- tests pass!


PyNite/Mesh.py
```diff
@@ -760,6 +761,39 @@ def __init__(self, x_left, y_bott, width, height):
         self.width = width
         self.height = height
 
+def determine_angle_offset(mesh:Mesh,axis,reassign=False):
```
JWock82, line 764: Will you help me understand what this is being used for?

PR #211, opened 2024-08-26
Vectorised Member Results Extraction

bjhowie:
The Member3D array results methods were just generating the results in a list comprehension then converting to a numpy array, which is quite slow. I have updated the methods to use vectorised numpy methods instead, unless a P-delta analysis has been conducted.

From the tests i've done, results extraction is approximately 10x faster for a member with 100 result stations.


bjhowie:
I suggest we update the tests to use Numpy < 2.0. Seems to be all that is failing.

connorferster:
@bjhowie Thanks for doing that. I did the original implementation of the results arrays methods so I am glad that you took the time to vectorize them...for a whopping 10X speed increase!!! That is awesome and will make such a difference.

connorferster (approved):

connorferster:
@JWock82 I will leave this PR for you to do a final review on in case there is something in the code that I missed during my review but I think it looks good.

PyNite/BeamSegZ.py
```diff
@@ -129,14 +130,18 @@ def axial(self, x):
...
+    def Torsion(self, x = 0):
         """
         Returns the torsional moment in the segment.
         """
         
         # The torsional moment is constant across the segment
         # This can be updated in the future for distributed torsional forces
-        return self.T1
+
+        if isinstance(x, (int, float)):
```
JWock82, line 141: I'm not sure what this change is doing. Can you provide some comments to clarify?

PyNite/Member3D.py
```diff
@@ -872,19 +872,32 @@ def shear_array(self, Direction, n_points: int, combo_name='Combo 1'):
             The number of points in the array to generate over the full length of the member.
         combo_name : string
             The name of the load combination to get the results for (not the load combination itself).
+        x_array : array = None
+            A custom array of x values that may be provided by the user, otherwise an array is generated.
+            Values must be provided in local member coordinates (between 0 and L) and be in ascending order
         """
         
-        # Segment the member into segments with mathematically continuous loads if not already done
```
JWock82, line 877: I'd like to have the documentation state why segmenting is necessary - to produce mathematically continuous beam segments (no point loads withing a segment). This allows us to use direct integration to obtain changes in moment, shear, slope and deflection inside the member. That may not be obvious to others.

PyNite/BeamSegZ.py
bjhowie: The reason for checking whether 'x' is a number or not is that these segment result functions can now be given an 'x' value that is actually an array of x values. The calling function expects the returned value to be an array of length equal to the input array of x values. For most functions, the returned value is calculated as a function of the input 'x' value which means the result is automatically equal in length to to 'x'. As the torsion is not calculated as a function of 'x', we need to manually check if 'x' is an array, and if so, return an array of equal length.

I am going to update these functions with type defintions, as a reminder in future that the function can accept either a number or an array of numbers.

PyNite/Member3D.py
bjhowie: Agreed. If i'm honest, I can't remember why this was changed. I'll revert.

PyNite/BeamSegZ.py
bjhowie: Actually scratch that last point. I'm not familiar with type annotations prior to python 3.10, which we would need to support. There is a broader conversation to be had there.

PR #213, opened 2024-09-02
Update FixedEndReactions.py

thisisapple:
The API itemset() not longer exists in Numpy V2.0. New users are all using Numpy V2.1. When they install PyNite and Numpy V2.1, an error will occur. To avoid the error, you must not use the itemset() method.

I change itemset() with this simple method:
Array[i, j] = your_value.
Anything else remains the same.


JWock82:
This has already been taken care of on the DKMQ branch, which will soon be merged with the main branch.

Issue #202, opened 2024-06-20
Moving to v1.0.0 w/ API consistency


connorferster:
**Is your feature request related to a problem? Please describe.**
PyNiteFEA continues to develop as a great project and I love using it (and teaching others to use it!) I know at least one engineer who relies on it for production engineering work (awesome!)

As PyNite is not in v1.0.0 yet, there have been some breaking changes to the API recently (part of which I addressed in #194). This is to be expected since the major version is still `0`. I am proposing that PyNite could be ready for a `v1.0` after doing some final house-keeping updates to the API. Once doing a v1.0 release, this would mean that the names of all classes, functions, parameters, and parameter order would stay fixed until it is decided that the existing API no longer serves the future of the project. 

To be clear, additional features can be added to the v1.0.0 release (e.g. new element types, new solvers, new methods, new attributes, etc.) but the default settings would need to remain and the new features would be optional.

(Apologies if you are nodding your head saying, "yes, yes, I know this..."; I am just stating it for clarity)

**Describe the solution you'd like**
I am proposing to open a PR (off of the DKMQ branch) to perform the following tasks to move towards a consistent API and to facilitate a stable v1.0 release:

* Make all attributes snake case. e.g. `model.Members` would become `model.members`, and `model.LoadCombos` would become `model.load_combos`
* Make all module names snake case. e.g. `import Rendering` would become `import rendering`. 
    * The exception to this would be FEModel3D.py which I think should remain `import FEModel3D` for historical reasons and because I think it just makes sense.
* Add type annotations to all user-facing function parameters and class attributes (or all functions/methods, if so desired)
    * @dinglind brought this up in #199, also. They make a good point and I would like to do this also. Type hints improve code completion and inline help prompts.
* Any other suggestions or changes you would like me to make.

**Describe alternatives you've considered**
We could also do nothing and leave things as they are. Breaking changes currently seem to happen unexpectedly though and, when they happen, it gives the impression that PyNite is "buggy". I think that PyNite is stable enough to consider a v1.0 release and give users confidence going forward that their scripts and models written against v1.x releases will always be able to run. This still allows PyNiteFEA to continue to grow rapidly through the addition of new features as attributes, parameters, functions, and classes can always be _added_ without having to change the names of existing attributes, function, and classes.

**Additional context**
Moving to a 1.0.0 release can also facilitate improved git branch management techniques. See #201

bjhowie:
[...]

PR #218, opened 2024-09-24
Enforce Use of Section Properties

bjhowie:
Following on from the discussion here #202 and working towards a v1.0.0 release, this pull request enforces the use of section properties when defining Member3D elements. This simplifies the processes of creating Member3D objects, by effectively grouping parameters to be passed to the `__init__ ` method. It also ensures consistency in the way that section properties are stored within the FEA model, which is critical as the package matures.

This pull request also alters the Member3D class logic slightly, so that the section and material properties are not extracted in the Member3D `__init__ ` method, but instead accessed directly from the appropriate section or material instances during analysis. This  enables simple alteration of section or material properties between analysis runs, with the confidence that any changes are propogated through to all members referencin gthe given section or material.

The same approach could be taken for Quad & Plate elements, but can be covered in a different pull request.

These changes will break most user's code.

### DIFF
--- a/Examples/Beam on Elastic Foundation.py
+++ b/Examples/Beam on Elastic Foundation.py
@@ -38,14 +38,16 @@ E = 29000   # ksi
 G = 11200   # ksi
 boef.add_material('Steel', E, G, 0.3, 490/1000/12**3)
 
-# Define section properties (W8x35)
+# Define section properties (W8x35) and add a section to the model
 A = 10.3    # in^2
 Iz = 127    # in^4 (strong axis)
 Iy = 42.6   # in^4 (weak axis)
 J = 0.769   # in^4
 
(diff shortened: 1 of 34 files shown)


KwentiN-ui:
I am not sure where exactly, but it seems the docstrings got lost for the section and material classes regarding this pull request 🤔

bjhowie:
> I am not sure where exactly, but it seems the docstrings got lost for the section and material classes regarding this pull request 🤔

I'm not sure if the Section or Material ever had docstrings to begin with. I added basic docstrings to both classes as part of this pull request.

JWock82 (approved):

JWock82:
Great work @bjhowie! I merged this pull request. This is going to force some longtime users (including myself) to adapt, but I think it's the direction `Pynite` needs to move. A very useful pull request!

PR #226, opened 2024-12-03
Switch to pytest framework for testing

### LINKED ISSUE #225: Switch to pytest framework
I propose switching to the 3rd party `pytest` framework from the builtin `unittest` framework.

The `pytest` framework has a number of advantages over `unittest`, and is very popular, used by major libraries such as NumPy. Critically, switching to pytest will not require rewriting any existing tests. Some benefits include:
* The ability to mark tests as slow and skip them for faster development.
* Easier to use, because it does not require that tests inherit from `TestCase`.
* Detailed and easy to understand failing test summaries.

This change will lower the barrier to writing tests for contributors and reduce the friction in running tests by providing an option to skip slow tests. I hope that these improvements will lead to more tests where appropriate and a more stable library.

I've successfully run pytest over the existing test suite, and would be happy to open a small PR upon consensus.

For more information see the docs: https://docs.pytest.org/en/stable/

jonbiemond:
Add the `pytest` library for running tests as detailed in issue #255

* Update `requirements.txt` to included `pytest`
* Run tests in the test-and-build workflow with `pytest` instead of `unittest`
* Add instructions to `CONTRIBUTING.md` on how to run tests with `pytest`

Resolves #225

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -40,9 +40,9 @@ jobs:
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
-    - name: Test with unittest
+    - name: Test with pytest
       run: |
-        xvfb-run -a coverage run --source=PyNite -m unittest discover
+        xvfb-run -a coverage run --source=PyNite -m pytest
 
     - name: Convert coverage to XML
       run: |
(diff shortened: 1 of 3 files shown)


JWock82:
This is great. I just want to be sure the test reports generated are ready to plug in to codecov before I pull it. I plan to have a closer look at it this next week.

JWock82:
I've looked over the changes you've made. All of the tests in the `Testing` folder will need to be changed from using `unittest` to `pytest` before I can merge this code.

PR #238, opened 2025-01-19
API consistency + Type annotations

bjhowie:
Some minor changes to the FEModel3D class for API consistency. Namely, `add_material` wasn't returning the new material name it had just created and the `add_section` methods didn't allow for automatic naming like the rest of the factory methods.

I have also added basic type annotations to the main user-facing methods of FEModel3D.

### DIFF
--- a/PyNite/FEModel3D.py
+++ b/PyNite/FEModel3D.py
@@ -1,4 +1,7 @@
 # %%
+#futures import required to use bar operators for optional type annotations
+from __future__ import annotations
+
 from os import rename
 import warnings
 from math import isclose
@@ -17,6 +20,7 @@ from PyNite.Plate3D import Plate3D
 from PyNite.LoadCombo import LoadCombo
 from PyNite.Mesh import Mesh, RectangleMesh, AnnulusMesh, FrustrumMesh, CylinderMesh
 from PyNite import Analysis
+
 
 # %%
 class FEModel3D():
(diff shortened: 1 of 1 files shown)


bjhowie:
Ah this has come up before.

May I suggest that in the interest of the vast majority of users, we consider dropping support for python 3.8 & 3.9 and start supporting 3.12 and 3.13 (assuming dependencies allow it).

JWock82:
It looks like these changes did not pass the python 3.8 and 3.9 checks. Here's the error I get:

image

bjhowie:
Yes, it's the bar operator in the type annotations. I'll switch them to the gross `typing.Optional` operator for backwards compatibility.

PyNite/FEModel3D.py
```diff
@@ -264,15 +281,25 @@ def add_steel_section(self, name:str, A:float, Iy:float, Iz:float, J:float,
...
+        # Name the section or check it doesn't already exist
+        if name:
+            if name in self.sections:
+                raise NameError(f"Section name '{name}' already exists")
         else:
-            # Stop execution and notify the user that the section name is already being used
-            raise Exception('Cross-section name ' + name + ' already exists in the model.')
-        
-    def add_spring(self, name, i_node, j_node, ks, tension_only=False, comp_only=False):
+            # As a guess, start with the length of the dictionary
```
JWock82, line 289: Am I looking at this wrong, or is the indentation off here

JWock82 (approved):

PR #252, opened 2025-04-07
Added type hints throughout

JBloss1517:
This add type hints to almost all class and method functions throughout library. I tried to scope new type related imports under the `if TYPE_CHECKING:` flag to minimize actual runtime performance hits from additional imports.

There were a few places where the types were a little ambiguous (mostly on the internal methods), so the types might be narrowed too much. For example, the argument x might permit a list of floats inl ieu of just a single float. 
 
Also, I Left type errors that show up in the editor after defining types on things largely untouched to keep scope of this PR strictly to adding type hints to the code. Future PRs can address these kind of warnings. These warnings were good to be warned about and should probably be fixed at some point to prevent errors popping up later.

I also largely left the imports as you have them (i.e. didn't remove unused imports), but I did try to sort them into PEP8 style order (standard lib -> 3rd party -> this package).

All tests are still passing.

### DIFF
--- a/Pynite/Spring3D.py
+++ b/Pynite/Spring3D.py
@@ -1,9 +1,17 @@
 # %%
+from __future__ import annotations # Allows more recent type hints features
 from numpy import zeros, array, add, subtract, matmul, insert, cross, divide
 from numpy.linalg import inv
 from math import isclose
 import Pynite.FixedEndReactions
 from Pynite.LoadCombo import LoadCombo
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Dict, Optional
+    from numpy import float64
+    from numpy.typing import NDArray
+    from Pynite.Node3D import Node3D
 
 # %%
 class Spring3D():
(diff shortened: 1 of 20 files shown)


Pynite/Spring3D.py
```diff
@@ -1,9 +1,17 @@
...
 from numpy.linalg import inv
 from math import isclose
 import Pynite.FixedEndReactions
 from Pynite.LoadCombo import LoadCombo
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Dict, Optional
+    from numpy import float64
+    from numpy.typing import NDArray
```
JWock82, line 13: `from typing import NDArray, Tuple`

JWock82:
Wow! This is a very helpful PR! Thanks for doing this! It looks the tests didn't pass due to an import error in `ShearWall.py`. I've commented with what I think should be the corrected code.

Pynite/Spring3D.py
JBloss1517: Thank you for reviewing this. You are right, I need to add Tuple to the `from typing import ...` line. I didn't run flake8 on the repo before pushing it up. But will be sure to run it locally before pushing up the change to make sure I didn't miss anything obvious.

JWock82:
@JBloss1517 you have been through a lot of the code on Pynite. Would you like to become a maintainer for Pynite?

KwentiN-ui:
I really appreciate this PR! It makes working with Pynite a lot more comfortable. Thank you for your time!

I've noticed that a bunch of files use conditional imports for types, which is a great way to get around circular imports. However using these types (especially in function definitions) can introduce a lot of headaches when not encapsulating them in quotations. I've whipped up a minimal example.

When running this file on Python 3.12, this will result in an import error:
```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from typing import Tuple

def foo(a:Tuple[int,int]):
    return a
```
This is a know problem with conditional type imports and can easily be avoided, by supplying the types as a string in quotation marks:
```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from typing import Tuple

def foo(a:"Tuple[int,int]"):
    return a
```
This still allows the LSP to parse the types correctly, without introducing weird behaviour.
I havent run into any issues yet, but this might be a useful thing to add to improve stability. If you don't mind the mixing of features I could also add the quotation marks to my PR #254.

I wish you all a nice weekend!

JBloss1517:
Thanks @KwentiN-ui for your kind words! I'm glad you find it helpful. When I was adding the type hints, I considered the string approach like you mentioned, but as strictly a matter of preference, I liked to keep them off. 

To avoid the runtime errors you mention you have two options:

1. wrap all type hints with quotes (this is what Python does under the hood anyway). Most editors are smart enough to understand what is going on and still work with them.

2. Add the following line at the top of the file: `from __future__ import annotations`. This allows you to not wrap your types in quotes and to also use more recent typing features like Union operators `|` instead of `Union[..., ...]` and lower case dict instead of Dict, etc. 

Ultimately I went with option 2 because it gained us some niceties and also avoids all the additional quotes. In Python 3.14 neither quotes nor the top __future__ import will be required 🎉, but it will be sometime before we can use this. 
You can check out https://peps.python.org/pep-0563/ for more background on this.

KwentiN-ui:
@JBloss1517 Thank you for the explanation. You obviously put some thought into the decision. I was not aware that `from __future__ import annotations` was this versatile, it seems to be pretty powerful regarding supporting older python versions.

JBloss1517:
> @JBloss1517 you have been through a lot of the code on Pynite. Would you like to become a maintainer for Pynite?

I would be honored to help in a more official capacity. My work (Apex Engineers) is currently using PyNite on some internal tooling and are willing to donate some of my time to help support this library for the broader community. I would probably stay clear of touching some of the matrix pieces until I have a better understanding of these subjects, but I can help in many other areas in the meantime.

PR #254, opened 2025-04-22
Writing FEModel3D Instances to the VTK format for postprocessing in ParaView

KwentiN-ui:
I'd like to contribute my `VTKWriter` Class to the Pynite project. It's purpose is to write solution data into the VTK format, so it can easily be postprocessed using [Paraview](https://www.paraview.org/) (see screenshots). This can be used as an alternative to the vtk rendering functionality, aimed at quick interactive reviews, aswell as powerful post-processing tools and raytraced renders.

This allows for a postprocessing workflow similar to solvers like OpenFOAM and Calculix. This does not add any additional dependencies, given that vtk is already used in the project.

image
example
example2

The `VTKWriter` can be instantiated with any `FEModel3D` instance. It supplies the `write_to_vtk` method to write a VTK file to disk at the specified path as well as the `open_in_paraview` that opens the file directly in paraview (if installed).

TODO List:
- [x] Nodes
    - [x] output external constraints
- [x] Member results like shear and bending moment
    - [x] basic member rendering and data output
    - [x] subdivide Members into many lines for accurate sampling of continous functions
- [x] ~~Plate3D~~ Relevant methods don't work with meshgrid, making it incredibly slow. Plates can simply be replaced by quads if needed 
- [x] Quad3D
    - [x] basic Quad rendering and data output
    - [x] use multiple vtkQuad Elements to sample the Pynite Quad at many points for improved accuracy
- [x] be able to preview the model before solving it, to check for obvious mistakes with the setup

I am looking forward to your feedback and opinions on this!

Example (see last line):
```python
from Pynite import FEModel3D
from Pynite.VTKWriter import VTKWriter

# Create a new model
model = FEModel3D()

# ... setup your model

# Analyze the frame
model.analyze()

# Write to VTK
VTKWriter(model).write_to_vtk("example")
# or open directly in ParaView (paraview needs to be installed and on system path), without explicitly saving first
VTKWriter(model).open_in_paraview()
```

### DIFF
--- a/Pynite/Node3D.py
+++ b/Pynite/Node3D.py
@@ -5,6 +5,7 @@ Created on Thu Nov  2 18:04:56 2017
 @author: D. Craig Brinck, SE
 """
 
+from __future__ import annotations
 from typing import Optional
 
 # %%      
--- a/Pynite/Node3D.py
+++ b/Pynite/Node3D.py
@@ -4,17 +4,16 @@ Created on Thu Nov  2 18:04:56 2017
 
 @author: D. Craig Brinck, SE
 """
-
-from __future__ import annotations
-from typing import Optional
-
+from __future__ import annotations # Allows more recent type hints features
 # %%      
+from typing import List, Tuple, Dict
+
 class Node3D():
     """
     A class representing a node in a 3D finite element model.
     """
     
-    def __init__(self, name, X, Y, Z):
+    def __init__(self, name: str, X: float, Y: float, Z: float):
         
         self.name = name                 # A unique name for the node assigned by the user
         self.ID: Optional[int] = None    # A unique index number for the node assigned by the program
@@ -23,52 +22,52 @@ class Node3D():
         self.Y = Y          # Global Y coordinate
         self.Z = Z          # Global Z coordinate
         
-        self.NodeLoads = [] # A list of loads applied to the node (Direction, P, case) or (Direction, M, case)
+        self.NodeLoads: List[Tuple[str, float, str]] = [] # A list of loads applied to the node (Direction, P, case) or (Direction, M, case)
         
         # Initialize the dictionaries of calculated node displacements
-        self.DX = {}
-        self.DY = {}
-        self.DZ = {}
-        self.RX = {}
-        self.RY = {}
-        self.RZ = {}
+        self.DX: Dict[str, float] = {}
+        self.DY: Dict[str, float] = {}
+        self.DZ: Dict[str, float] = {}
+        self.RX: Dict[str, float] = {}
+        self.RY: Dict[str, float] = {}
+        self.RZ: Dict[str, float] = {}
         
         # Initialize the dictionaries of calculated node reactions
-        self.RxnFX = {}
-        self.RxnFY = {}
-        self.RxnFZ = {}
-        self.RxnMX = {}
-        self.RxnMY = {}
-        self.RxnMZ = {}
+        self.RxnFX: Dict[str, float] = {}
+        self.RxnFY: Dict[str, float] = {}
+        self.RxnFZ: Dict[str, float] = {}
+        self.RxnMX: Dict[str, float] = {}
+        self.RxnMY: Dict[str, float] = {}
+        self.RxnMZ: Dict[str, float] = {}
 
         # Initialize all support conditions to `False`
-        self.support_DX = False
-        self.support_DY = False
-        self.support_DZ = False
-        self.support_RX = False
-        self.support_RY = False
-        self.support_RZ = False
+        self.support_DX: bool = False
+        self.support_DY: bool = False
+        self.support_DZ: bool = False
+        self.support_RX: bool = False
+        self.support_RY: bool = False
+        self.support_RZ: bool = False
 
         # Inititialize all support springs
-        self.spring_DX = [None, None, None]  # [stiffness, direction, active]
-        self.spring_DY = [None, None, None]
-        self.spring_DZ = [None, None, None]
-        self.spring_RX = [None, None, None]
-        self.spring_RY = [None, None, None]
-        self.spring_RZ = [None, None, None]
+        self.spring_DX: List[float | str | bool | None] = [None, None, None]  # [stiffness, direction, active]
+        self.spring_DY: List[float | str | bool | None] = [None, None, None]
+        self.spring_DZ: List[float | str | bool | None] = [None, None, None]
+        self.spring_RX: List[float | str | bool | None] = [None, None, None]
+        self.spring_RY: List[float | str | bool | None] = [None, None, None]
+        self.spring_RZ: List[float | str | bool | None] = [None, None, None]
 
         # Initialize all enforced displacements to `None`
-        self.EnforcedDX = None
-        self.EnforcedDY = None
-        self.EnforcedDZ = None
-        self.EnforcedRX = None
-        self.EnforcedRY = None
-        self.EnforcedRZ = None
+        self.EnforcedDX: float | None = None
+        self.EnforcedDY: float | None = None
+        self.EnforcedDZ: float | None = None
+        self.EnforcedRX: float | None = None
+        self.EnforcedRY: float | None = None
+        self.EnforcedRZ: float | None = None
 
         # Initialize the color contour value for the node. This will be used for contour smoothing.
-        self.contour = []
+        self.contour: List[float] = []
 
-    def distance(self, other):
+    def distance(self, other: 'Node3D') -> float:
         """
         Returns the distance to another node.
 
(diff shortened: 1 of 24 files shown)


KwentiN-ui:
I just noticed the other pull request #252 that added typehints already. All my changes in other files aside from the `VTKWriter` file can be overwritten with their changes without problem.

KwentiN-ui:
Seems like I need to get rid of the match statements, should be an easy fix 👍

KwentiN-ui:
@JBloss1517 I am glad the code was easy to follow! About your questions:
1. I have taken a look at the `Visualization.py` module beforehand, given that it's using `vtk` package aswell. Unfortunately they don't have a lot of overlapping code, as it focuses more on the rendering aspect (i.e. Annotations, Lines, etc.). The VTKWriters primary goal is to simply fetch all the "raw" data that's there and write it to disk. Everything related to Visualization is then done by Paraview using filters.
2. This is definetly a concern of mine too. As the Writer only fetches data from the Model, the correctness should be covered by other tests specifically targeting solution accuracy. My main focus of the unittest was to catch python exceptions thrown on writing to avoid obvious issues on code changes. However, it was necessary to write some custom data fetching code that would really be suited inside the actual class and be tested seperately (like outputting Quad3D data as a 2D Meshgrid instead of only on specific points). Though as this is my first PR here I didn't really want to touch the other files. My suggestion would be to open up a new PR where we could discuss/implement these first, update this PR to use the new methods directly from the quad and member instances, than merge.
3. I used the `test_2D_frames.py` file as a template and didn't think much more of it. I am not very experienced with unit testing but using the `pytest` command seemed to work fine with it. I'll have a second look though 👍🏻

I know this PR might come a bit out of the blue for anyone who hasn't used Paraview before but I encourage you to try it out if you have some spare time. It's a great piece of FOSS that also uses Python as a scripting language, so there is a fair bit of overlap there.

JWock82:
I am vaguely familiar with Paraview. I have used it minimally with OpenSees models. Many scientific Python users are familiar with Paraview, so I think exporting to Paraview is a feature worth adding. I am unfamiliar with the process of exporting to Paraview myself, but I think as long as the code is contained in its own module that doesn't affect any of the other modules in Pynite this shouldn't be a problem.

Some of the tests do test the ability of the VTK code to run in a headless (no screen) environment. You may be able to use those as a starting point. Early on I was using unittest to write the tests, but I recently switched to pytest. You can use either and pytest will read it, but the preferred method going forward is pytest rather than unittest.

I'm sorry I've been a bit absent from Pynite lately. My work situation has changed recently, and I've been trying to ensure I have an agreement with my employer that keeps my work on Pynite open-source going forward. This morning I received word that I have permission to continue developing Pynite as an open-source project, so we're all good now.

Pynite/Node3D.py
```diff
@@ -15,8 +15,8 @@ class Node3D():
     
     def __init__(self, name: str, X: float, Y: float, Z: float):
         
-        self.name = name    # A unique name for the node assigned by the user
-        self.ID = None      # A unique index number for the node assigned by the program
+        self.name = name                 # A unique name for the node assigned by the user
+        self.ID: Optional[int] = None    # A unique index number for the node assigned by the program
```
JWock82, line 19: The node ID should never be optional. It is assigned by the program once analysis starts. This is the program's way of internally tracking all the degrees of freedom on the model. Having an ID separate from the name allows the user to use whatever name they want without interfering with the analysis.

JWock82:
I've examined the code and appreciate this pull request. Integration with Paraview is a feature that many users will want.

It looks like the build failed when I ran it. Have a look at the reasons the checks were not successful. It looks like it has to do with the changes made to node.ID initialization, which is "mission critical" for Pynite to run.

Pynite/Node3D.py
KwentiN-ui: Hello @JWock82 ! Thank you for taking a look at my code. The `Optional[int]` is a type hint that tells the user that the value can be `None` at some point during execution. Typehints do not effect code execution in any way.  `Optional` is a so called `Union` type that contains both the `int` and the `None` type. So it could also be written as `self.ID: int|None`, where `|` is the OR statement. The name `Optional` just means that it might exist, depending on the state of the analysis. I made a small example to show why this is useful:
grafik

The build error resulted from a missing import statement, propably related to me merging this file with the other typehint changes upstream. I fixed it just now in the latest commit.

KwentiN-ui:
Can you try running the build tools again? Should be working now ^^

JBloss1517:
I am still learning the matrix math part of quads and plates, so I apologize if I am missing your question. But what I think you are referring to is the same problem noted in this issue #192. @JWock82 I believe you have already taken a look at it in #192. Did you come up with something that @KwentiN-ui can use or build upon? The ability to easily plot quad data in a similar manner as we do in the Member3D seems like a really nice addition.

Regarding the branching, you can branch off your main branch and try to get something working, and then when it feels like you have it solved sufficiently, you can merge that branch back into your main branch, and this PR will be updated. That will allow you to develop the quad plotting feature without spamming this PR with lots of commits that may or may not pan out.

KwentiN-ui:
@JBloss1517 It's more about passing a grid of points to get all values at once over calling the quads method multiple times, which then recalculates a bunch of the same data between subsequent runs.

If you want to, for example, plot a surface f(x,y) using numpy, you'd use the same approach.
```python
import numpy as np
# first setup x and y as simple vectors, both of shape (50,1)
x,y = [np.linspace(0,1,50)]*2
# create a coordinate grid, making both their shapes to (50,50)
X,Y = np.meshgrid(x,y)
# compute the function f, for every point on the grid at once
F = np.sin(X) + np.cos(Y) # still shape (50,50)

# plot the function as a contour plot looking from above
import matplotlib.pyplot as plt
plt.contourf(X,Y,F, levels=20)
plt.show()
```
image

Applying this principle to quads would mean to have dedicated methods for the quad, that can take in these coordinate grids and spit out an [m,n] grid of values sampled from the entire area of the quad. Coincidentally this just works as is for the `Quad3D`s functions if local coordinates are used, but I can't expect them to work forever. Also this means that I need to handle the conversion from local->global inside the writer, which is not ideal given that the Quad can already handle this for single points.

The coordinate grid for the following example would look like this spelled out:
quad

$$
\xi= \begin{bmatrix}
0 & 0.25 & 0.5 & 0.75 & 1 \\
0 & 0.25 & 0.5 & 0.75 & 1 \\
0 & 0.25 & 0.5 & 0.75 & 1 \\
0 & 0.25 & 0.5 & 0.75 & 1 \\
0 & 0.25 & 0.5 & 0.75 & 1 \\
\end{bmatrix}, \eta=\begin{bmatrix}
0 & 0 & 0 & 0 & 0 \\
0.25 & 0.25 & 0.25 & 0.25 & 0.25 \\
0.5 & 0.5 & 0.5 & 0.5 & 0.5 \\
0.75 & 0.75 & 0.75 & 0.75 & 0.75 \\
1 & 1 & 1 & 1 & 1
\end{bmatrix}
$$

This already exists for `Member3D`. The methods end on `_array` and take in the number of points to sample and spit out the vector of size n sampling the members data along it's length. This would be the 1D case for this exact usecase. The quad version could look like this: `shear_array(n_points:Tuple[int,int], local:bool, combo_name:str) -> np.ndarray`

JWock82:
FYI: the domain/range of eta and xi are (-1, 1) rather than (0, 1)

JWock82:
> I am still learning the matrix math part of quads and plates, so I apologize if I am missing your question. But what I think you are referring to is the same problem noted in this issue #192. @JWock82 I believe you have already taken a look at it in #192. Did you come up with something that @KwentiN-ui can use or build upon? The ability to easily plot quad data in a similar manner as we do in the Member3D seems like a really nice addition.
> 
> Regarding the branching, you can branch off your main branch and try to get something working, and then when it feels like you have it solved sufficiently, you can merge that branch back into your main branch, and this PR will be updated. That will allow you to develop the quad plotting feature without spamming this PR with lots of commits that may or may not pan out.

Issue #192 is a different issue. @KwentiN-ui is asking for a way to calculate results for multiple points on a plate at the same time. Currently you must calculate the results at one point at a time, leading to some repeat calculations that could avoided when multiple points are calculated for a plate.

JWock82:
```
> # first setup x and y as simple vectors, both of shape (50,1)
> x,y = [np.linspace(0,1,50)]*2
> # create a coordinate grid, making both their shapes to (50,50)
> X,Y = np.meshgrid(x,y)
```

FYI: the domain/range of eta and xi are (-1, 1) rather than (0, 1). The natural coordinate system's origin is at the center of the element, rather than the corner.

KwentiN-ui:
@JWock82 Oh thats really important, I'll fix it right away! Thank you for pointing this out.

PR #259, opened 2025-05-12
test: added continuous beam test to check max moment results

connorferster:
Quickly implemented test to validate the results of the `FEModel3D.moment_array` method.

Will add additional tests for the other array methods which will be used to validate that `max(model.something_array(...)[1]) == model.max_something(...)` and `min(model.something_array(...)[1]) == model.min_something(...)`.

These tests are expected to fail on the current version of PyNite at the time this pull request was opened.

### DIFF
--- a/Testing/test_continuous_beam.py
+++ b/Testing/test_continuous_beam.py
@@ -1,0 +1,47 @@
+import unittest
+from Pynite import FEModel3D
+import math
+import sys
+from io import StringIO
+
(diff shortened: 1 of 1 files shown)


JWock82:
I've added the array functions to physical members and I've pulled these tests in. We'll se how they run in a moment, and then I'll fix any remaining bugs in the next few days. Once everything's proven I'll issue a new version 1.1.0 with the patch and a few new features as well.

Issue #260, opened 2025-05-14
add_member_self_weight() needs clarification.


dankicode:
**Describe the bug**
Per the docstring, `FEModel3D.add_member_self_weight()` expects a `global_direction`, either 'FX', 'FY', or 'FZ'. 
It then uses `add_member_dist_load()` to add the self weight. However, there is no validation of the global direction, meaning that local directions ('Fx', 'Fy', or 'Fz') may still be passed and runs without errors.

**To Reproduce**
See last sentence above.

**Expected behavior**
1. A value error should be raised if a local direction is provided. 
2. In addition, I believe it would be more clear if a default value of `global_direction: str = 'FZ'` and `factor: float = -1.0` is provided.

@JWock82 - Let me know if you are fine with the above and I'd be happy to issue a PR. Happy to do just item 1 above if you'd like to not provide the default parameters.

JWock82:
I have merged the code you've provided. Sorry that I didn't see it before I released v1.1.0. It will be included with the next release.

PR #262, opened 2025-05-15
Add validation to reject local directions in add_member_self_weight

dankicode:
Bug is described in Issue #260.


JWock82:
Thanks for adding this validation code!

PR #280, opened 2025-09-08
feat: implement dynamic __version__

### LINKED ISSUE #273: `__version__` should be populated in `__init__.py`
**Is your feature request related to a problem? Please describe.**
Without a variable called `__version__` in the `__init__.py` file, it is hard to check which version of Pynite is installed. Being able to check the version installed is an important step in debugging.

**Describe the solution you'd like**
Add `__version__` to `__init__.py`.

I will do a PR for this.

connorferster:
Version is retrieved dynamically from the package after it is built by setup.py. This allows users to access the version of Pynite they have installed without @JWock82 having to manually update the version number in `__init__.py` whenever a new release is made.

Closes #273 

In the long-term, I recommend moving away from setup.py to pyproject.toml where the version and description of the package can be dynamically retrieved from the `__init__.py` file (making the `__init__.py` file the source of truth and the package being built according to those values). 

However, this is a quick fix that will allow users to easily check their Pynite version to aid in the diagnosis of potential incorrect values.

### DIFF
--- a/Pynite/__init__.py
+++ b/Pynite/__init__.py
@@ -1,3 +1,12 @@
 # Select libraries that will be imported into Pynite for the user
 from Pynite.FEModel3D import FEModel3D
 from Pynite.ShearWall import ShearWall
+import Pynite
+
+from pip._vendor import pkg_resources
+
+def get_version(package):
+    package = package.lower()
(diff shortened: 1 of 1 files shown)


JWock82:
Thanks for this @connorferster

PR #281, opened 2025-09-08
tests: add test that fails for Pynite 1.2.0, 1.3.0

connorferster:
**Draft**

There was a regression error introduced into Pynite 1.2.0 that has persisted into 1.3.0. I am not sure the exact cause but I can describe the symptoms:

1. In the statics check, the sum of the moments do not equal `0.0`
2. The moment diagrams are incorrect

It _appears_ to me that the fixed-end moments are coming through as the member moments without the joint releases being applied.

Here is a script that causes this issue:

```python
  model = FEModel3D()
  
  model.add_material("default", 1, 1, 1, 1, 1)
  model.add_section("default", 1, 1, 1, 1)
  
  model.add_node("0", 0, 0, 0)
  model.add_node("1", 10, 0, 0)
  model.add_node("2", 13, 0, 0)
  
  model.def_support("0", 1, 1, 1, 1, 1, 0)
  model.def_support("1", 0, 1, 0, 0, 0, 0)
  
  model.add_member("M0", "0", "2", "default", "default")
  
  model.add_member_dist_load("M0", "Fy", -10, -10, 0, 13, case='load')
  
  model.add_load_combo("combo", {"load": 1.0})
  
  model.analyze(check_statics=True)
  
  model.members['M0'].plot_moment("Mz", "combo", 2000)
  
  model.members['M0'].max_moment("Mz", "combo")
```

Here is what the results look like on **Pynite 1.2.0 & 1.3.0**:

image

And here is how things look on **Pynite 1.1.2**:

image

**I have opened this pull-request to contribute the code above as a failing test.** However, this PR does not provide a fix.

### DIFF
--- a/Testing/test_continuous_beam.py
+++ b/Testing/test_continuous_beam.py
@@ -2,8 +2,7 @@ from Pynite import FEModel3D
 import math
 # import matplotlib, matplotlib.pyplot as plt
 
-def create_model():
-
+def create_3_support_model():
     continuous_beam = FEModel3D()
 
     continuous_beam.add_node('N1', 0, 0, 0)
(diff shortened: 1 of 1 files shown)


connorferster:
> I've figured out what's going on. This was subtle. Starting in v1.2.0 I've been using Flake8 to help me conform to the PEP8 style guide for python. Flake8 flags boolean comparisons such as `if my_bool == True` and suggests replacing them with `if my_bool is True`. I've been doing that. An unintended consequence is that your use of `model.def_support('1', 1, 1, 1, 1, 1, 0)` gets treated incorrectly. When I switched your code to `model.def_support('1', True, True, True, True, True, False)` everything works again. The diagrams it was producing for you were the fixed end reactions for the beam segments.

Hmmm... I am not so sure that is it. I have tried that also because I noticed a strange behaviour where `1 != True` before.

This still creates the same problem for me:

```python
model = FEModel3D()

model.add_material("default", 1, 1, 1, 1, 1)
model.add_section("default", 1, 1, 1, 1)

model.add_node("0", 0, 0, 0)
model.add_node("1", 10, 0, 0)
model.add_node("2", 13, 0, 0)

model.def_support("0", True, True, True, True, True, 0)
model.def_support("1", 0, True, 0, 0, 0, 0)

model.add_member("M0", "0", "2", "default", "default")

model.add_member_dist_load("M0", "Fy", -10, -10, 0, 13, case='load')

model.add_load_combo("combo", {"load": 1.0})

model.analyze(check_statics=True)

model.members['M0'].plot_moment("Mz", "combo", 2000)
model.members['M0'].max_moment("Mz", "combo")
```

image

connorferster:
@JWock82 My bad. In the example above, I only changed the `1`s to `True`s. If I _also_ change the `0`s to `False`s, then it works.

Would you be open to a PR where I change the check for fixities back to an equality check from the `is` operator? I will add the `#noqa` comments to the lines so that Flake8 does not object during CI/CD.

JWock82:
I would prefer Pynite to use the more robust solution. I'd like to switch any text ` is True` and ` is False` to ` == True` and ` == False`. I didn't realize it would cause an issue like this.

JWock82:
I can turn off the Flake8 rule on my end. No need to add additional comments.

JWock82:
I'll take care of this and update it. You don't need to do anything. This is a simple find and replace operation in VS Code.

PR #288, opened 2025-11-04
Add modal analysis feature to PyNite

boustrephon:
Sorry about the delay in getting this to you. There have been a few little tweaks around the place, so please let me know if you are not happy with the approach. The first one is probably the one you need to focus on. I can take a different approach if you prefer.

I have:
- Added reference to FEModel3D in each Node3D  ** this is a breaking change that was done to make load combinations easily accessible and to bring the Nodes3D class into line with Member3D **
- Added “analyze_modal”, which generates a mass matrix from the matrices of members and nodes.
- Added mass generation for FEModel3D, which calls on mass generation for members and nodes
- Added mass matrix generation for members and nodes based on self weight and / or load combo
- Added “lumped_mass” option to members
- Added reference to FEModel3D (self.model) to definition of node so that load combinations can be easily referenced when defining masses based on load combination.
- Added an example file
- Added a test file

### DIFF
--- a/Pynite/Analysis.py
+++ b/Pynite/Analysis.py
@@ -528,6 +528,25 @@ def _sum_displacements(model: FEModel3D, Delta_D1: NDArray[float64], Delta_D2: N
         node.RZ[combo.name] += Delta_D[node.ID*6 + 5, 0]
 
 
+def _expand_displacements(model, D1, D2, D1_indices, D2_indices):
+    """
+    Expands partitioned displacement vectors back to full DOF set.
+    Used for mode shape expansion.
+    """
+    total_dof = len(model.nodes) * 6
+    D_full = zeros((total_dof, 1))
+    
+    # Place D1 values (unknown DOFs)
+    for i, index in enumerate(D1_indices):
+        D_full[index] = D1[i]
+    
+    # Place D2 values (known DOFs - typically zeros for modal analysis)
+    for i, index in enumerate(D2_indices):
+        D_full[index] = D2[i]
+    
+    return D_full
+
+
 def _check_TC_convergence(model: FEModel3D, combo_name: str = "Combo 1", log: bool = True, spring_tolerance: float = 0, member_tolerance: float = 0) -> bool:
     """Checks for convergence in tension-only and compression-only analysis.
 
--- a/Pynite/Node3D.py
+++ b/Pynite/Node3D.py
@@ -6,7 +6,18 @@ Created on Thu Nov  2 18:04:56 2017
 """
 from __future__ import annotations  # Allows more recent type hints features
 # %%
-from typing import List, Tuple, Dict, Optional
+from numpy import array, zeros
+
+from typing import List, Tuple, Dict, Optional,TYPE_CHECKING
+if TYPE_CHECKING:
+
+    from typing import Dict, List, Tuple, Optional, Any, Literal
+
+    from numpy import float64
+    from numpy.typing import NDArray
+
+    from Pynite.FEModel3D import FEModel3D
+    from Pynite.LoadCombo import LoadCombo
 
 
 class Node3D():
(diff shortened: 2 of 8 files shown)


boustrephon:
I have not added any visualisation (rendering).

My main concern is that the modal frequencies do not match the OpenSees example all that closely, especially with the higher modes.

JWock82:
Thanks for contributing this! I'll review it as soon as I can and either comment on it or pull it into the main branch. It appears all the tests have passed, which indicates nothing got seriously broken. I've got one change in line ahead of this before pulling it in.

boustrephon:
I think the only potential issue is if anyone has existing code that instantiates a Node3d object directly.

Pynite/Node3D.py
```diff
@@ -78,3 +94,106 @@ def distance(self, other: 'Node3D') -> float:
...
+
+    def M(self, mass_combo_name: str = '', 
+        mass_combo_direction: int = 2, characteristic_length: float|None = None) -> NDArray:
+        """Returns the node's mass matrix (6x6 diagonal). In member-based models, nodes provide
+           translational mass only to prevent double-counting of rotational inertia. For member-less
+           models, rotational inertia can be added by providing a characteristic_length.
+
+        :param mass_combo_name: Load combination name for force-based mass calculation, defaults to ""
+        :type mass_combo_name: str, optional
+        :param mass_combo_direction: Direction for mass conversion: 0=X, 1=Y, 2=Z, defaults to 2
```
JWock82: 'Y' is the default vertical axis in `Pynite`. Recommend defaulting to 1.

Pynite/Analysis.py
```diff
@@ -524,6 +528,25 @@ def _sum_displacements(model: FEModel3D, Delta_D1: NDArray[float64], Delta_D2: N
...
+    # Place D1 values (unknown DOFs)
+    for i, index in enumerate(D1_indices):
+        D_full[index] = D1[i]
+    
+    # Place D2 values (known DOFs - typically zeros for modal analysis)
+    for i, index in enumerate(D2_indices):
+        D_full[index] = D2[i]
+    
+    return D_full
+
```
JWock82: Redundant: There is already a function in Analysis.py that does this. Have a look at _unpartition_disp.

boustrephon:
Great. Thank you for all your work on fixing things and getting it to align with your code. 

I have done a fair amount of Python programming over the last ten years (and a lot of structural analysis over more than thirty years), but this was the first time I have ever approached FEA. 

I used DeepSeek to propose code based on your existing code, which seemed to work quite well. I tested the code against some theoretical cantilever modes and I had to interact with DeepSeek quite a lot, asking questions and making sure that I understood it. It certainly wasn't a case of simple "vibe-coding", but it would have taken me a lot longer on my own, and DeepSeek seems to have done a good job of following your style.

I will take another look at it shortly. I am thinking about response spectrum analysis.

I note that you have code for splitting members at nodes. How well does that work?

Incidentally, I have some code for parsing ETABS text files. Would you be interested in an ETABS converter?

JWock82:
I've been finding AI is a helpful tool too. I definitely am cautious with what it produces, as I don't want Pynite to turn into a giant "vibe-coded" mess. I'm glad you have a background in Python to start with and are reviewing the AI output carefully. I reviewed the code you submitted extensively myself, and also had many discussions with AI, and dove into my textbooks to understand better what it was doing. I think I have a solid understanding of modal analysis now. I have made quite a few changes to your code to simplify choices for the user and to make it fit into Pynite's existing framework a little better.

Incidentally, I now have mode shapes rendering. Instead of storing them in a `modal_results` dictionary, I store the mode shape directly in the nodes like any other displacement. I have the analysis engine add a load combination for each mode named 'Mode 1', 'Mode 2', etc. You can simply render those combinations to see the mode shapes now (with loads turned off for now - there's an exception there in the rendering engine I've been meaning to address). I've gotten rid of the `modal_results` dicitonary, but I've kept `frequencies` as a model attribute for easy access.

I would be interested in response spectrum analysis. I'd like to finish cleaning up modal analysis. There are endless features to add or improve upon. I don't use ETABS myself, but I'm not opposed to building out features other users find useful. I need more collaborators since I am the bottleneck. Would you be interested in being added as a collaborator?

PR #289, opened 2025-11-13
Changing member to Timoshenko beam

jrrifareal:
Changes:
- Stiffness matrix of Member3D was modified to account for shear deformation. A derivation is included.
- Fixed-end reactions were also modified. Derivations were also included.
- Section was modified to accept optional arguments for shear correction factors

Everything still defaults to Euler-Bernoulli beam unless shear_deformable=True is set (see example). The stiffness matrix and fixed-end reactions reduce to Euler-Bernoulli beam if shear_deformable=False.

If you do not like the idea of directly changing the formulas in Member3D, I have created another version which has a new class that inherits to Member3D and just overrides the stiffness and fixed-end reaction calculation. 

Thanks.


boustrephon:
> I’ve altered it in the code already and I’ll get it corrected in the example too before the next release. I’m working through all the modal analysis code one more time and reformatting it to be consistent with the rest of Pynite’s conventions. I should note that I removed the M trace output from the logs. From what I can tell, that output is meaningless when a consistent mass matrix is used.

Yes, sorry about leaving that trace in. I thought I had removed it.

JWock82:
No apologies necessary. I appreciate your contribution. It’s very helpful. Something I’ve wanted to add for a long time but just didn’t have time to prioritize. I’m glad you found time to do it.

JWock82:
@jrrifareal I've modified your code here to match some recent changes I've made. It looks like you've added/modified ~1400 lines of code. That's awesome, but it will take some serious time to review. A lot of the modified code is pretty central to Pynite's core, so I want to get it right. I've created a `shear_deformation` branch in Pynite that I am pulling your changes into. I'll use it to review your changes over time. I'll keep it up to date with other changes going on in main. Once I'm satisfied everything is correct, I'll merge it into the main branch. Thanks for this contribution!

PR #291, opened 2025-11-16
Drop support for Python 3.7

### LINKED ISSUE #290: Drop support for Python 3.7
I'd like to propose replacing the `pip` dependency but it requires a builtin feature not available before Python 3.8. Python version 3.7 reached [end-of-life more than two years ago](https://devguide.python.org/versions/). Can we move to require `python >= 3.8`?

Currently the oldest version of Python receiving updates is 3.10, however I see no problem supporting allowing installs on 3.8 and 3.9 till we require features from a more recent version.

Alternatively, if support for 3.7 is desired, than we can use an explicitly declared dependency.

jonbiemond:
This allows the project to leverage Python features introduce in Python 3.8.
Python 3.7 reached its [end-of-life on June 27, 2023](https://peps.python.org/pep-0537/).

Resolves #290

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,5 +33,5 @@ setuptools.setup(
         'derivations': ['jupyterlab', 'sympy']
     },
     include_package_data = True,
-    python_requires = '>=3.7',
+    python_requires = '>=3.8',
 )


JWock82 (changes requested):
I agree with @JBloss1517. Let's move this all the way to 3.10.

PR #297, opened 2025-12-16
Add pyupgrade with pre-commit

ulgens:
Modernizes the codebase using [pyupgrade](https://github.com/asottile/pyupgrade). This was a very easy win on linting, but I can improve the setup by adding ruff and other common tools in following PRs.


JBloss1517:
I love this PR, thanks for submitting it! Now that we are in python 3.10, moving away from the extra imports for things like list, dict, tuple, etc. is great. It looks like the extra imports are still at the top of each file though. Can you remove any unused imports that were created by this upgrade?

PR #299, opened 2025-12-17
Check unused imports via ruff

ulgens:
Adds a very basic ruff config to check F401 - unused import rule.

The ruff setup can be extended later.

I added a simple workflow to run pre-commit hooks in CI.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,14 @@
+name: lint
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  prek:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+      - uses: j178/prek-action@v2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.14.9
+    hooks:
+      - id: ruff-check
+        args: [--fix, --show-fixes]
--- a/Examples/Circular Bin with Conical Hopper.py
+++ b/Examples/Circular Bin with Conical Hopper.py
@@ -1,6 +1,4 @@
-from math import pi
 from Pynite.FEModel3D import FEModel3D
-from Pynite.Mesh import FrustrumMesh, CylinderMesh
 
 t = 0.25/12
 E = 29000*1000*12**2
--- a/Examples/Pushover Analysis.py
+++ b/Examples/Pushover Analysis.py
@@ -1,4 +1,3 @@
-from math import isclose
 
 from Pynite.FEModel3D import FEModel3D
 
--- a/Examples/Simple Beam - Factored Envelope.py
+++ b/Examples/Simple Beam - Factored Envelope.py
@@ -2,11 +2,8 @@
 
 # Import `FEModel3D` from `Pynite`
 from Pynite import FEModel3D
-from matplotlib import pyplot as plt
-import numpy as np
 
 # Import 'Visualization' for rendering the model
-from Pynite import Visualization
 
 # Create a new finite element model
 simple_beam = FEModel3D()
--- a/Examples/Simple Beam - Point Load.py
+++ b/Examples/Simple Beam - Point Load.py
@@ -5,7 +5,6 @@
 from Pynite import FEModel3D
 
 # Import 'Visualization' for rendering the model
-from Pynite import Visualization
 
 # Create a new finite element model
 simple_beam = FEModel3D()
--- a/Pynite/Analysis.py
+++ b/Pynite/Analysis.py
@@ -4,7 +4,7 @@ from math import isclose
 
 import warnings
 
-from numpy import array, atleast_2d, zeros, subtract, matmul, divide, seterr, nanmax, asarray, isfinite
+from numpy import array, zeros, subtract, matmul, asarray, isfinite
 from numpy.linalg import solve, norm, LinAlgError
 
 from Pynite.LoadCombo import LoadCombo
@@ -1495,7 +1495,8 @@ def _partition_D(model: FEModel3D) -> Tuple[List[int], List[int], NDArray[float6
             D2_indices.append(node.ID*6 + 5)
             D2.append(0.0)
 
-    # Legacy code on the next line. I will leave it here until the line that follows has been proven over time.
+    # Legacy code on the next lines. I will leave it here until the line that follows has been proven over time.
+    # from numpy import atleast_2d
     # D2 = atleast_2d(D2)
 
     # Convert D2 from a list to a matrix
--- a/Pynite/FEModel3D.py
+++ b/Pynite/FEModel3D.py
@@ -5,7 +5,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Literal
 
 import numpy as np
-from numpy.linalg import solve
 import scipy as sp
 
 from Pynite.Node3D import Node3D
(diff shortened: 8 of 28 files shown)


ulgens:
Rebased and ready for review.

JBloss1517:
Thanks for following up and keeping this PR alive! I can get it reviewed this week.

jonbiemond (commented):
Awesome to see! Amazing how many unused imports stick around.

I just want to note that by running ruff via prek we're missing out on inline annotations in GitHub. See the workflow examples provided by Astral: https://docs.astral.sh/ruff/integrations/#github-actions

Of course, in the case maintainers or contributors choose to adopt prek locally, the benefit of using prek in CI is that the lint config is specified in only one location.

ulgens:
I agree the inline annotations looks nice, but they are limited to certain tools and normally the expectation is that the developers will fix these issues in their local and CI check is there just to ensure they are running it locally.

JBloss1517:
Thanks for kicking off the ruff stack — nice to see #299 → #300 laddering up the ruleset incrementally. Keeping this PR to F401 only is the right call; my notes are all small/config-level so the foundation grows cleanly:

**Worth fixing here:**

1. **`target-version` mismatch** — `ruff.toml` sets `py310`, but `setup.py` requires `>=3.11` (since #291 dropped 3.7). Match them so ruff (and the `UP` rules coming in #300) target the right baseline:
   ```toml
   target-version = "py311"
   ```
2. **Align excludes with `.flake8`** — `.flake8` already skips `Derivations`, `docs`, `Resources`, `.Archived`, `build`, `dist`. Porting those now keeps noise down as the ruleset expands in #300:
   ```toml
   extend-exclude = ["*.ipynb", ".Archived", "Derivations", "docs", "Resources", "build", "dist"]
   ```
3. **Drop `--fix` from the committed hook** — the same `.pre-commit-config.yaml` drives local commits and CI. In CI, `--fix` edits files that then get discarded and the failure reads as "files were modified by this hook" instead of a clean violation list. Let the shared config report; devs run `ruff check --fix` locally:
   ```yaml
   hooks:
     - id: ruff-check   # report only
   ```

**On @jonbiemond's annotation point** — agreed, and since ruff is currently the only check, I'd lean toward Astral's native `ruff` action (`--output-format=github`) for this job to get inline PR annotations now, and revisit prek-in-CI once the hook set grows beyond ruff — that's when the single-config-location benefit actually pays for the lost annotations. Either way config stays in `ruff.toml`, so it's reversible.

**Optional:** bump `j178/prek-action@v1` → `@v2` (checksum validation + caching) if you stay on prek; rename the `Prek checks` workflow to `lint`; and align its trigger (`[push, pull_request]`, all branches) with `build-and-test.yml`'s `branches: [main]`.

**All good as-is** — the `__all__` addition correctly marks the re-exports as intentional for F401, and the commented-out import notes are a nice touch. No need to touch flake8 here; see #327 for the overall direction.

JWock82:
I agree with @JBloss1517's assessment. @ulgens, if you address @JBloss1517's comments I'll merge the pull request. Thanks for your work on this!

ulgens:
> target-version mismatch

Fixed.

---

> Align excludes with .flake8

`build` and `dist` are already parts of [the default ruff config](https://docs.astral.sh/ruff/configuration/), I added the rest.

---

> Drop --fix from the committed hook — the same .pre-commit-config.yaml drives local commits and CI. In CI, --fix edits files that then get discarded and the failure reads as "files were modified by this hook" instead of a clean violation list. Let the shared config report; devs run ruff check --fix locally

I believe that the auto-fix feature is useful for the development process, and personally, I use it a lot to auto-apply the fixes. This is also the common behaviour among the git hooks.

That being said, I agree that a non-descriptive message in CI can be improved. I think this can be solved by `--show-fixes`:

```
      --show-fixes
          Show an enumeration of all fixed lint violations. Use
          `--no-show-fixes` to disable
```

https://docs.astral.sh/ruff/configuration/#full-command-line-interface

I added `--show-fixes` to the ruff args but if you think this doesn't produce the result you are looking for, I can remove both of the args.

On a personal note; not all git hooks provide this feature and my take is that for this specific setup, it's more important to show that the workflow is failing than showing the exact details of why it's failing. If we assume that the expectation from contributors is to run it (automatically) before committing their changes and the CI is here to check that. In case of a failure, the only thing they need to do is to do run the hooks locally - they are not expected to observe and fix individual issues.

---

About the annotations vs git hooks case,

I think the git hooks setup can be enriched a lot beyond ruff. [This](https://github.com/ulgens/django-blasphemy/blob/c14cd83842aa80935d9fe2fe027b2039d88beb11/.pre-commit-config.yaml) is the standard config I use for the projects I work on. Afaik, except zizmor, none of the other tools provide this annotation feature and I value the uniformity of the reports, so I lean towards the git hooks option.

---

Applied the rest of the recommendations.

JBloss1517:
Verified and good to merge from my side.

I merged this against current `main` locally, since the base is about six weeks
behind now, and re-ran everything:

- `ruff check .` is clean. Bare `main` has 48 F401s, so the fix set is still
  complete and nothing new crept in from the commits since the merge-base.
- 18 of 22 `Pynite` submodules import cleanly. The other four (`Rendering`,
  `Reporting`, `VTKWriter`, `Visualization`) fail on optional deps I don't have
  installed locally, not on anything this PR touches.
- pytest, with the three render-dependent test files excluded for the same
  reason: 127 passed, 1 skipped, 18 subtests passed, 1 failed. The failure is
  `test_circular_hopper`, which dies importing `Pynite/Visualization.py`, a file
  this PR doesn't modify. CI covers that path properly and is green.

I also went through the removals that could plausibly bite rather than just
trusting the green checks. The one I expected to be a problem, dropping the bare
`import Pynite.FixedEndReactions` from `Spring3D.py`, is fine: `Member3D.py`
imports it directly and is the only real consumer, so that copy was redundant.
Dropping `from IPython.display import Image` in `Rendering.py` is a small bonus
on top, since it takes a hard IPython dependency off importing that module. Good
call adding the `# from numpy import atleast_2d` note above the commented-out
legacy line in `Analysis.py`.

On `--fix` and `--show-fixes`: agreed, keep it as you have it. `--show-fixes`
covers what I was actually worried about, which was an opaque CI failure, and
every F401 is auto-fixable, so the enumeration is complete in practice. Easy
enough to revisit if it reads badly once #300 lands.

On prek versus the native `ruff` action, you've convinced me and I'll withdraw
the suggestion. My position was "native action now, revisit prek once the hook
set grows beyond ruff." But #327 is aiming at uv + ruff + pyrefly, so that
growth is already on the board. Adopting the native action now would mean
ripping it back out in a few weeks, which isn't worth it just for annotations in
the meantime.

One gap I'll pick up separately, not something for you to fix here: nothing
currently tells contributors any of this exists. `requirements.txt` lists only
`pytest` under `# Development`, and `CONTRIBUTING.md` has a `## Testing` section
but says nothing about linting. So someone can clone, commit, and get a red
`lint` check from a tool they were never pointed at, with no local hook
installed. That's the premise your `--fix` reasoning rests on, so it's worth
making real. I'll add `ruff` and `prek` to the dev block and a short `## Linting`
section after `## Testing`. I'll fold in two cosmetic bits at the same time: the
now-dangling `# Import 'Visualization' for rendering the model` comments in the
two `Examples/Simple Beam - *.py` files, and `"Pynite"` in `__all__`, which is
only there to silence F401 on a self-import that looks vestigial.

None of that blocks this. @JWock82, good to go from my end.

Thanks for keeping this alive since December and for working through the config
notes.

ulgens:
(rebased over the latest main)

JWock82:
@JBloss1517 do I need to take any action here? Or are we waiting on other parts of the roadmap described in #327 to be resolved first?

JBloss1517:
@JWock82 this one can be merged. I'm always hesitant to merge into main, which is why I haven't just merged it yet. Our contributing docs already help people know they will need to install prek outside of the repo (if they don't have it). Once prek is available it should auto pickup the pre commit configuration this PR includes to make sure we don't leave unused imports laying around going forward.

PR #301, opened 2025-12-19
Run examples as part of test suite

### LINKED ISSUE #294: Run examples as part of test suite
The examples can be run during tests to surface errors. This **does not** include making sure the results are correct. This adds another layer of checks to catching any breaking changes to the public API.

A downside is that it lengthens the time it takes to run the test suite. They take approximately 2 minutes and 12 seconds on my machine. This could be addressed by excluding the tests by default, and including them as part of CI.

Let me know and I will open a PR.

jonbiemond:
Run all files in [/Examples](https://github.com/JWock82/Pynite/tree/main/Examples) as a test.

Resolves #294

### DIFF
--- a/Testing/test_examples.py
+++ b/Testing/test_examples.py
@@ -1,0 +1,29 @@
+import glob
+import importlib
+from unittest.mock import Mock
+
+import pytest
+
+PROJECT_ROOT = importlib.resources.files("Pynite").parent
+EXAMPLES_DIR = PROJECT_ROOT / "Examples"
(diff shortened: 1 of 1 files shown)


jonbiemond:
As mentioned in #294, I don't love that these add +2 minutes to the time it takes to run tests. One option is to skip them by default while still including them in CI.

jonbiemond:
There are a few examples that are still broken. A couple fixes are waiting in PRs #311 and #326. And one still needs to be fixed as reported in issue #324.

jonbiemond:
This PR is stuck until all examples are working.

PR #311, opened 2026-03-10
Fix Moment Frame example

jonbiemond:
Regression introduced by d302d46
Update the argument to the new name.


jonbiemond:
@JWock82 Could you please take another look at this when you have a chance? It is blocking #301

PR #315, opened 2026-03-30
Clarify the member local coordinate system convention

bjude:
Updates the documentation to reflect the logic in `Member3D.T()`

There is also a drive-by fix to a relevant section in rendering.rst, a missing newline broke the rendering of the member local coordinate system section

### DIFF
--- a/docs/source/member.rst
+++ b/docs/source/member.rst
@@ -44,11 +44,23 @@ Local Coordinate System
 Each member starts at its i-node and ends at its j-node. The local x-axis for the member is defined
 by a vector going from the i-node to the j-node.
 
-By default the local z-axis is always horizontal in Pynite, and is on the right-hand side of the
-member.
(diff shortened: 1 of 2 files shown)


JWock82 (changes requested):

JWock82:
I've reviewed the sign conventions in detail, and agree with Copilot's assessment. I think changing the documentation too much will cause confusion. Much of the confusion comes from whether we are talking about bending "along" an axis or "about" an axis. Either way, the local coordinate systems are rendering properly. I've stopped referring to the direction as strong and weak.

bjude:
I disagree that its confusing (for one with any engineering background) to refer to strong/weak axis, but if theres different terminology that you think would make things clearer/more consistent across the project then i'll happily adopt that.

Axis naming aside, the core purpose of this PR is documenting the default member orientation and Pynite's default definition of "Up", which is currently undocumented. Its a footgun that should be clarified, even if this PR isnt the right way to do it.

Issue #275, opened 2025-07-09
Model produces results when unstable with a hinge


weikequ:
**Describe the bug**
Define a pin supported beam with a hinge in the middle. Analyzing the structure using `analyze()` and it will run properl, but give errorneous reactions, moments, shears, and displacements. Additionally, running using `analyze(sparse=False)` will give a different set of results.

See video below.

**To Reproduce**
Here's the code to reproduce the behaviour:

```python
from Pynite import FEModel3D

model = FEModel3D()

Iy = 0.000015234  # m^4 weak axis
Iz = 0.0000712  # m^4 strong axis
J = 0.000000241  # m^4
A = 0.006264504  # m^2
model.add_section("W10x33", A, Iy, Iz, J)

E = 200000000  # kPa
G = 77000000  # kPa
nu = 0.3  # Poisson's ratio
rho = 77  # Density (kN/m^3)
model.add_material("steel", E, G, nu, rho)

model.add_node("n1", 0, 0, 0)
model.add_node("n2", 1, 0, 0)
model.add_node("n3", 2, 0, 0)

model.add_member("m1", "n1", "n2", "steel", "W10x33")
model.add_member("m2", "n2", "n3", "steel", "W10x33")

model.def_support("n1", 1, 1, 1, 1, 1, 0)
model.def_support("n3", 0, 1, 1, 1, 1, 0)

model.add_member_dist_load("m1", "Fy", 10, 10)
model.add_member_dist_load("m2", "Fy", 10, 10)

model.def_releases("m2", Rzi=True)

model.analyze(sparse=False)

print(model.nodes["n1"].RxnFY)
print(model.nodes["n3"].RxnFY)
print(model.nodes["n2"].DY)
```

**Expected behavior**
An error should be thrown about the model being unstable.

**Screenshots**

https://github.com/user-attachments/assets/18006125-fadd-4904-8b9f-f83b719af06c

angelmusonda:
Hi @weikequ,

You've raised an important issue, which you also brought up in #255. @JWock82 mentioned that he would take a closer look.

I'd like to offer a few comments.

Firstly, PyNite internally uses the "is" operator for Boolean comparisons instead of "==". This means that using 0 and 1 in methods that expect True and False may not behave as expected. The "is" operator checks for object identity, not value equality. As a result, in your model, all degrees of freedom at nodes n1 and n3 were fully restrained. However, this is beside the point, because even if you explicitly write:

```
[...]

PR #320, opened 2026-06-01
Detect global instability via the solution residual (fixes #255, #275)

### LINKED ISSUE #255: No error is thrown on unstable model in some scenarios
**Describe the bug**
In some cases, when using `FEModel3D.analyze()`, when the model is obviously unstable, no errors are thrown. This happens with the default options, but with `sparse=False`, errors are sometimes correctly thrown. I've come across a couple different scenarios that I'll go through below.
- Scenario 1: A warning is sent to `stderr`: `MatrixRankWarning: Matrix is exactly singular D1 = spsolve(K11.tocsr(), subtract(subtract(P1, FER1), K12.tocsr() @ D2))`, but no error is thrown and code execution continues.
- Scenario 2: No warnings are given, no errors are thrown, code execution continues.
- Scenario 3: Same as Scenario 2, but no errors are thrown when even `sparse=False` is added

**To Reproduce**
Scenario 1:
```python
from Pynite import FEModel3D
m = FEModel3D()
m.add_node("N1", 0, 0, 0)
m.add_node("N2", 0, 1, 0)
m.add_material("Steel", 29000, 11400, 0.5, 490 / 1000 / 12**3)
# No supports are added
m.add_section("Section", 10, 100, 150, 250)
m.add_member("M1", "N1", "N2", "Steel", "Section")
m.analyze()
```

Scenario 2:
```python
from Pynite import FEModel3D
m = FEModel3D()
m.add_node("N1", 0, 0, 0)
m.add_node("N2", 0, 12 * 12, 0)
m.add_node("N3", 15 * 12, 12 * 12, 0)
m.add_node("N4", 15 * 12, 0 * 12, 0)
m.add_material("Steel", 29000, 11400, 0.5, 490 / 1000 / 12**3)
m.add_section("Section", 10, 100, 150, 250)
m.add_member("M1", "N1", "N2", "Steel", "Section")
m.add_member("M2", "N4", "N3", "Steel", "Section")
m.add_section("Section2", 15, 100, 250, 250)
m.add_member("M3", "N2", "N3", "Steel", "Section2")
m.add_node_load("N2", "FX", 50)
# No supports are added
m.analyze()
```

Scenario 3:
```python
from Pynite import FEModel3D
m = FEModel3D()
m.add_node("N1", 0.0, 0.0, 0.0)
m.add_node("N2", 4.0 , 0.0, 0.0)
m.add_node("N3", 4.0 , 3.0 , 0.0)
m.add_node("N4", 0.0, 3.0 , 0.0)
m.add_material("Steel", E=29000, G=11200, nu=0.3, rho=0.00028356481481481483)
m.add_section("W10x33", A=9.71, Iy=36.6, Iz=171, J=0.58)
m.add_member("M1", "N1", "N2", "Steel", "W10x33")
m.add_member("M2", "N2", "N3", "Steel", "W10x33")
m.add_member("M3", "N3", "N4", "Steel", "W10x33")
m.add_member("M4", "N4", "N1", "Steel", "W10x33")
m.add_node_load("N2", "FX", 50)
# No supports are added
m.analyze(sparse=False)
```
**Expected behavior**
An exception should be thrown if we encounter any singular matrices/there is any global instability.

gaoflow:
## Summary

Fixes #255 and #275. A globally **unstable** model — a mechanism (e.g. a beam with an internal hinge and no support at the hinge) or a structure with insufficient supports (rigid-body motion) — has a *singular* global stiffness matrix even though every diagonal term is non-zero. `analyze()` currently solves it anyway and silently returns erroneous reactions, shears, moments and displacements instead of raising.

The existing nodal stability check (`Analysis._check_stability`) only inspects the diagonal for unsupported zero-stiffness DOFs, so it cannot see this kind of *global* rank deficiency. The underlying solvers don't reliably report it either:

- `scipy.sparse.linalg.spsolve` returns `NaN` (with only a `MatrixRankWarning`), or in some cases finite-but-meaningless displacements with **no warning at all** (#255 scenario 2).
- `numpy.linalg.solve` can miss the singularity due to round-off (#255 scenario 3), and neither solver flags the hinge mechanism of #275.

## Approach

I add `Analysis._solve_unknown_disp(K11, rhs, sparse, check_stability)`, which solves for the unknown displacements and, when `check_stability` is enabled, **verifies the solution against equilibrium** using the relative residual `‖K11·x − rhs‖ / ‖rhs‖`. A stable structure — even a numerically ill-conditioned one — solves to within round-off because the direct solvers are backward-stable, whereas a singular system yields a non-finite or O(1)-residual result.

I first tried the natural pivot-magnitude / rank approach (factorize and compare the smallest LU pivot to the largest). It is **not sound for this codebase**: several stable models in the existing test suite (e.g. the Kassimali 2D frame, and the mat-foundation / plate meshes) have a smallest-to-largest pivot ratio already down at ≈1e-17, indistinguishable from a genuine mechanism, so any pivot threshold misclassifies them. Measured across the suite, the worst stable relative *residual* is ≈2.4e-13, while the unstable models from #255/#275 give residuals of order 1 — a clean ~13-order-of-magnitude separation. The default tolerance is `1e-6`.

The check is cheap (one matrix–vector product plus two norms, no extra factorization) and gated on the existing `check_stability` flag, so the historical fast path is preserved when it is disabled.

## Scope

The first-order elastic / tension-compression path (`Analysis._first_order`, used by `analyze()`) and the linear path (`FEModel3D.analyze_linear`) now route through the helper. The P-Delta and pushover paths are intentionally left unchanged — they deliberately drive the structure toward buckling (where the tangent stiffness approaches singular by design), so a hard instability raise there would need separate, buckling-aware handling.

## Reproduction & verification

All four reported scenarios now raise `The stiffness matrix is singular ...` instead of returning results, for both the sparse and dense solvers:

- #275: simply-supported two-span beam joined by an internal hinge with no support at the hinge.
- #255 scenario 1: a single unsupported member.
- #255 scenario 2: an unsupported frame loaded in-plane (previously returned finite garbage with no warning).
- #255 scenario 3: an unsupported closed frame solved with `sparse=False`.

Added `Testing/test_instability.py` covering the hinge mechanism and the unsupported-model scenarios (sparse and dense), that a stable propped cantilever still solves to the correct reaction (`5wL/8`), and that `check_stability=False` bypasses the check.

The full core test suite passes (the optional-dependency test files for the visualization/VTK backends fail to *collect* locally for unrelated reasons and were excluded): **121 passed** (115 existing + 6 new), with no regressions.

### DIFF
--- a/Pynite/Analysis.py
+++ b/Pynite/Analysis.py
@@ -2,8 +2,10 @@ from __future__ import annotations  # Allows more recent type hints features
 from typing import TYPE_CHECKING
 from math import isclose
 
-from numpy import array, atleast_2d, zeros, subtract, matmul, divide, seterr, nanmax
-from numpy.linalg import solve
+import warnings
+
+from numpy import array, atleast_2d, zeros, subtract, matmul, divide, seterr, nanmax, asarray, isfinite
+from numpy.linalg import solve, norm
 
 from Pynite.LoadCombo import LoadCombo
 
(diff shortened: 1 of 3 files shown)


JWock82:
Thanks for this excellent contribution. I made a few improvements to this with the help of GitHub Copilot and merged it all in PR #323.

binggao1230:
Thanks for folding it into #323 — glad the global-instability detection was useful, and the Copilot polish on top sounds great. Happy to see it land.

PR #326, opened 2026-06-07
Fix Pushover Analysis Example

### LINKED ISSUE #325: Pushover Analysis Example Fails
The [Pushover Analysis](...) example raises an exception evaluating the trace while analyzing the pushover. Unfortunately, the exception is suppressed so the exact error is unknown without more digging. This is a regression introduced in commit d19c3c8079866569d997f9a756a81543d1a8c1d6.

This issue is blocking PR #301.

Error: `RuntimeError: Error evaluating pushover trace 'Fixed End Moment' for load combination 'Primary'.`
<details>
  <summary>Full traceback</summary>

```bash
self = FEModel3D(solution='Pushover'), log = True, check_stability = False, push_combo = 'Pushover', max_iter = 30, tol = 0.01, sparse = True, combo_tags = None, control_node = 'N2', control_direction = 'DY'
control_limit = 6.0, traces = {'Fixed End Moment': <function <lambda> at 0x14b6df1c0>, 'Load Point Deflection': <function <lambda> at 0x14b6df320>, 'Load Point Moment': <function <lambda> at 0x14b6df060>}

    def analyze_pushover(self, log=False, check_stability=True, push_combo='Push', max_iter=30, tol=0.01, sparse=True, combo_tags=None, control_node=None, control_direction='DX', control_limit=None, traces=None):

        if control_limit is not None and control_node is None:
            raise ValueError('A control node must be specified when a pushover control limit is provided.')

        if control_direction.upper() not in ('DX', 'DY', 'DZ', 'RX', 'RY', 'RZ'):
            raise ValueError("Pushover control direction must be one of 'DX', 'DY', 'DZ', 'RX', 'RY', or 'RZ'.")

        if control_node is not None and control_node not in self.nodes:
            raise ValueError(f"Control node '{control_node}' was not found in the model.")

        if traces is not None:
            if not isinstance(traces, dict):
                raise ValueError('Pushover traces must be provided as a dictionary of trace names to callables.')

            for trace_name, trace in traces.items():
                if not callable(trace):
                    raise ValueError(
                        f"Pushover trace '{trace_name}' is not callable. Wrap expressions in a lambda or function that accepts combo_name."
                    )

        if log:
            print('+---------------------+')
            print('| Analyzing: Pushover |')
            print('+---------------------+')

        # Prepare the model for analysis
        Analysis._prepare_model(self)

        # Get the auxiliary list used to determine how the matrices will be partitioned
        D1_indices, D2_indices, D2 = Analysis._partition_D(self)

        # Identify and tag the primary load combinations the pushover load will be added to
        for combo in self.load_combos.values():

            # No need to tag the pushover combo
            if combo.name != push_combo:

                # Add 'primary' to the combo's tags if it's not already there
                if combo.combo_tags is None:
                    combo.combo_tags = ['primary']
                elif 'primary' not in combo.combo_tags:
                    combo.combo_tags.append('primary')

        # Identify which load combinations have the tags the user has given
        combo_list = Analysis._identify_combos(self, combo_tags)
        combo_list = [combo for combo in combo_list if combo.name != push_combo]

        # Initialize per-combo pushover result state.
        self._pushover_state = {}
        self._pushover_traces = {}

        # Step through each load combination
        for combo in combo_list:

            if log:
                print('')
                print('- Analyzing load combination ' + combo.name)

            # Member end forces will be summed across multiple load steps
            # Set member end force summations to zero
            for phys_member in self.members.values():
                for sub_member in phys_member.sub_members.values():
                    sub_member._fxi, sub_member._fyi, sub_member._fzi = {}, {}, {}
                    sub_member._mxi, sub_member._myi, sub_member._mzi = {}, {}, {}
                    sub_member._fxj, sub_member._fyj, sub_member._fzj = {}, {}, {}
                    sub_member._mxj, sub_member._myj, sub_member._mzj = {}, {}, {}

            # Get the partitioned global fixed end reaction vector for the load combination
            FER1, FER2 = Analysis._partition(self, self.FER(combo.name), D1_indices, D2_indices)

            # Get the partitioned global nodal force vector for the load combination
            P1, P2 = Analysis._partition(self, self.P(combo.name), D1_indices, D2_indices)

            # Run an elastic P-Delta analysis for the load combination (w/o pushover loads)
            # This will be used to preload the member with non-pushover loads prior to pushover anlaysis
            Analysis._PDelta(self, combo.name, P1, FER1, D1_indices, D2_indices, D2, False, sparse, check_stability, 30)

            # Seed the nonlinear end-force history with the elastic preload state from the
            # primary load combination before any pushover increments are applied.
            for phys_member in self.members.values():
                for sub_member in phys_member.sub_members.values():
                    f = sub_member.f(combo.name)
                    sub_member._fxi[combo.name] = f[0, 0]
                    sub_member._fyi[combo.name] = f[1, 0]
                    sub_member._fzi[combo.name] = f[2, 0]
                    sub_member._mxi[combo.name] = f[3, 0]
                    sub_member._myi[combo.name] = f[4, 0]
                    sub_member._mzi[combo.name] = f[5, 0]
                    sub_member._fxj[combo.name] = f[6, 0]
                    sub_member._fyj[combo.name] = f[7, 0]
                    sub_member._fzj[combo.name] = f[8, 0]
                    sub_member._mxj[combo.name] = f[9, 0]
                    sub_member._myj[combo.name] = f[10, 0]
                    sub_member._mzj[combo.name] = f[11, 0]

            # The P-Delta analysis above flagged the solution as a P-Delta solution, but we need to indicate that this is actually a Pushover solution so that the calls to Member3D.f() going forward are executed considering nonlinear behavior
            self.solution = 'Pushover'

            # Define the pushover load step and initialize the pushover load factor
            load_step = list(self.load_combos[push_combo].factors.values())[0]
            step_num = 1
            load_factor = load_step*step_num

            # Get the partitioned global fixed end reaction vector for one pushover increment.
            # The pushover combo factor already defines the increment size.
            FER1_push, FER2_push = Analysis._partition(self, self.FER(push_combo), D1_indices, D2_indices)

            # Get the partitioned global nodal force vector for one pushover increment.
            P1_push, P2_push = Analysis._partition(self, self.P(push_combo), D1_indices, D2_indices)

            self._pushover_traces[combo.name] = {
                trace_name: [] for trace_name in (traces or {})
            }

            self._pushover_state[combo.name] = {
                'status': 'running',
                'step_num': 0,
                'load_factor': 0.0,
                'control_node': control_node,
                'control_direction': control_direction.upper(),
                'control_limit': control_limit,
                'control_displacement': None,
                'message': None,
            }

            # Apply the pushover load in steps, summing deformations as we go, until the full pushover load has been analyzed
            while round(load_factor, 8) <= 1.0:

                # Inform the user which pushover load step we're on
                if log:
                    print('- Beginning pushover load step #' + str(step_num))
                    print(f'- Load_factor = {load_factor}')

                # Run the next pushover load step
                # Note: The validity of the pushover step is checked and handled within the _pushover_step method
                Analysis._pushover_step(self, combo.name, push_combo, step_num, P1_push, FER1_push, FER2_push, D1_indices, D2_indices, D2, log, sparse, check_stability, tol)

                control_displacement = None
                if control_node is not None:
                    node = self.nodes[control_node]
                    control_displacement = float(getattr(node, control_direction.upper())[combo.name])

                if traces is not None:
                    for trace_name, trace in traces.items():
                        try:
                            self._pushover_traces[combo.name][trace_name].append(trace(combo.name))
                        except Exception as exc:
>                           raise RuntimeError(
                                f"Error evaluating pushover trace '{trace_name}' for load combination '{combo.name}'."
                            ) from exc
E                           RuntimeError: Error evaluating pushover trace 'Fixed End Moment' for load combination 'Primary'.

Pynite/FEModel3D.py:2881: RuntimeError
```

</details>

jonbiemond:
Pass `plastic_beam` in via a default argument to avoid a `NameError` when called in another module during analysis.

Lambda functions only have access to the scope of the environment they are called in. Any additional variables from where they are defined must be explicitly passed in.

Resolves #325

### DIFF
--- a/Examples/Moment Frame - Pushover.py
+++ b/Examples/Moment Frame - Pushover.py
@@ -82,9 +82,9 @@ frame.add_load_combo('Pushover', {'Push': 0.01})
 # combo results for each load combo and load step. Traces are an optional feature for pushover
 # analysis intended to help visualize the progression of key response quantities.
 traces = {
-    'Node d Drift': lambda combo_name: frame.nodes['d'].DX[combo_name]*12,
-    'Moment at c:': lambda combo_name: frame.members['bd'].moment('Mz', x=20.0, combo_name=combo_name),
-    'Moment at d:': lambda combo_name: frame.members['bd'].moment('Mz', x=60.0, combo_name=combo_name),
+    'Node d Drift': lambda combo_name, frame=frame: frame.nodes['d'].DX[combo_name]*12,
+    'Moment at c:': lambda combo_name, frame=frame: frame.members['bd'].moment('Mz', x=20.0, combo_name=combo_name),
+    'Moment at d:': lambda combo_name, frame=frame: frame.members['bd'].moment('Mz', x=60.0, combo_name=combo_name),
 }
 
 # Analyze the model
--- a/Examples/Pushover Analysis.py
+++ b/Examples/Pushover Analysis.py
@@ -67,9 +67,9 @@ control_limit = -6.0
 # combo results for each load combo and load step. Traces are an optional feature for pushover
 # analysis intended to help visualize the progression of key response quantities.
 traces = {
-    'Fixed End Moment': lambda combo_name: plastic_beam.members['M1'].moment('Mz', x=0.0, combo_name=combo_name),
-    'Load Point Moment': lambda combo_name: plastic_beam.members['M1'].moment('Mz', x=96.0, combo_name=combo_name),
-    'Load Point Deflection': lambda combo_name: plastic_beam.nodes['N2'].DY[combo_name],
+    'Fixed End Moment': lambda combo_name, plastic_beam=plastic_beam: plastic_beam.members['M1'].moment('Mz', x=0.0, combo_name=combo_name),
+    'Load Point Moment': lambda combo_name, plastic_beam=plastic_beam: plastic_beam.members['M1'].moment('Mz', x=96.0, combo_name=combo_name),
+    'Load Point Deflection': lambda combo_name, plastic_beam=plastic_beam: plastic_beam.nodes['N2'].DY[combo_name],
 }
 
 # Analyze the model


JWock82:
Thanks for the contribution. I can see why this approach is useful in your case, and I don’t think there’s anything wrong with it technically.

For this example, though, I’d like to keep the code as simple and approachable as possible. Most new users won’t be calling the example from another module, so adding the explicit `frame=frame` argument introduces an extra concept that may create more confusion than clarity.

Because this seems specific to your workflow rather than something broadly needed by most users, I’m hesitant to incorporate it into the general example problem. I appreciate you pointing it out, though.

jonbiemond:
Hi @JWock82, sorry for not explaining correctly. The example does not work as is, unless pasted directly into a library file and ran from there.

PR #329, opened 2026-06-23
Update GHA dependencies

ulgens:
Updates outdated external GHA dependencies

Two notes:
* These type of dependency updates can be automated with [Renovate](https://docs.renovatebot.com/)
* Both of the existing workflows are missing some common security practices. Those can be detected and fixed by [zizmor](https://docs.zizmor.sh/)


JWock82:
Thanks for updating this!

PR #332, opened 2026-08-03
Document lint setup for contributors

JBloss1517:
Follow-up to #299, and **should merge after it** — the `## Linting` section references `ruff.toml` and `.pre-commit-config.yaml`, which that PR adds.

#299 adds a `lint` CI check, but nothing tells contributors it exists. `requirements.txt` listed only `pytest` under `# Development`, and CONTRIBUTING.md had no linting section at all. As it stands someone can clone, commit, and get a red check from tooling they were never pointed at, with no local hook installed.

This adds:

- `ruff` and `prek` to the development requirements
- A `## Linting` section after `## Testing` covering the hook install, running ruff directly, and a note that stock `pre-commit` works against the same config for anyone who already uses it

I verified every command against a tree with #299 merged in: `prek install` writes the hook, `ruff check --fix .` picks up `ruff.toml`, and committing a file with an unused import triggers the hook, which blocks the commit and reports the fix it made.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,3 +21,29 @@ Run tests using `pytest`.
 ```bash
 pytest
 ```
+
+## Linting
+
+Linting is handled by [ruff](https://docs.astral.sh/ruff/), with hooks managed by [prek](https://github.com/j178/prek). The rules live in `ruff.toml`, and currently only check for unused imports (F401).
+
+Install prek outside the project environment, so that git GUI clients can run the hooks without activating a virtualenv:
+
+```bash
+uv tool install prek
+```
+
+`pipx install prek` works too. Then install the git hook:
+
+```bash
+prek install
+```
+
+The configured checks now run against your staged files on every commit, fixing what they can. To run them across the whole repository:
+
+```bash
+prek run --all-files
+```
+
+The same checks run in CI on every pull request.
+
+prek reads `.pre-commit-config.yaml`, which is the standard pre-commit config format, so stock `pre-commit` works against the same config if you already use it. If you want ruff in your editor, install it separately and match the version pinned in `.pre-commit-config.yaml` so it reports the same results as the hook.


CONTRIBUTING.md
```diff
@@ -21,3 +21,24 @@ Run tests using `pytest`.
...
+pip install -r requirements.txt
+prek install
+```
+
+With the hook installed, ruff runs against your staged files on every commit and fixes what it can. The same checks run in CI on every pull request, so it's worth having the hook in place before you push.
+
+To run the checks yourself without the hook:
+
+```bash
+ruff check --fix .
```
ulgens: I'd recommend `prek run` or `prek run --all-files` here to keep it compatible in future with potential new tools in git hooks config.

requirements.txt
```diff
@@ -11,3 +11,5 @@ ipywidgets
 
 # Development
 pytest
+ruff
```
ulgens: If https://github.com/JWock82/Pynite/pull/332/changes#r3710318815 recommends using prek/pre-commit, this is not needed for local manual runs (but can still be useful for IDE integrations. In that case, I'd recommend pinning the version.)

requirements.txt
```diff
@@ -11,3 +11,5 @@ ipywidgets
...
+prek
```
ulgens: I prefer installing it system wide so git UIs/GUIs don't need to activate the environment to run the hooks.

JBloss1517:
Thanks for these, all three are going in.

I should say up front that I'm new to prek and pre-commit generally, so this is genuinely useful rather than a style preference I'm politely conceding. At least one of these I would not have worked out on my own.

On `prek run --all-files` over `ruff check --fix .`: agreed, and the reasoning clicked once I thought past ruff. My version only invokes ruff, so as soon as the hook config grows a second tool, the documented manual command quietly becomes a partial check while still looking like the full one. I checked that `prek run --all-files` works without running `prek install` first, so it stands on its own. I've dropped the `ruff check --fix .` line rather than keep both.

On `ruff` in requirements.txt: dropped. The pinning point is what convinced me. prek installs its own ruff at `v0.14.9` from `.pre-commit-config.yaml`, so an unpinned `ruff` sitting in requirements.txt would drift from that and report differently in an editor than on commit. Rather than carry a second version to keep in sync, I've added a line to the docs telling anyone who wants editor integration to match the pinned rev.

On installing prek system wide: this is the one I'd missed entirely, and it's a real failure mode rather than a preference. I hadn't considered that hooks fired from a git GUI don't activate the virtualenv, so a venv-only prek would just silently not run. I've switched the docs to `uv tool install prek`, with pipx mentioned as an alternative, and confirmed it lands on PATH outside the project environment.

Net effect is that requirements.txt drops out of this PR entirely, so it's now CONTRIBUTING.md only. Your three comments deleted half the PR and improved it, which I appreciate.

JWock82:
@copilot please review this pull request for me. Also, flag any security vulnerabilities you see for me.

Issue #327, opened 2026-06-15
Umbrella: consolidate onto a uv + ruff + pyrefly toolchain


JBloss1517:
### Goal
  
Move Pynite dev tooling to a cohesive, fast, mostly-Rust toolchain. 

Here is what I'm proposing to add: 
- uv (packaging)
- ruff (lint/format)
- pyrefly or ty (type checking)
- possibly add ruff format so formatting on files becomes consistent across the whole project

This would replace the current flake8 + ruff overlap, ad-hoc pip installs, and absence of type checking in CI/development outside of editor type checking.

Here are some of the existing PRs or issues that are related: 
- #299 — ruff F401 + pre-commit + lint workflow (first step, intentionally minimal)
- #300 — adds UP (pyupgrade) ruleset; depends on #299, supersedes #297
- #297 (closed) — standalone pyupgrade, replaced by ruff approach
- #298 (closed) — ruff format trial: large diff due to "scientific"/matrix styling — formatting deferred/contentious
- #252 (merged) — type hints already present throughout

### Proposed workstreams (need maintainers buy-in)
1. Packaging → uv + pyproject.toml. Migrate setup.py to pyproject.toml; map extras_require to dependency groups; add uv.lock; swap the CI pip block for setup-uv + uv sync.
 2. Linting → finish ruff rollout. Continue stacked PRs after #300 (E, W, I), migrate .flake8 ignores into ruff.toml, then retire flake8 (delete .flake8, drop the flake8 step in build-and-test.yml).
 3. Type checking → pyrefly. Stable v1.0 (May 2026); proven on PyTorch/JAX (numerics-heavy, like us). Start lenient, tighten over time. Wire into CI (`uv run pyrefly check`) and pre-commit.
   - Alternatives considered: ty (Astral; pairs with ruff/uv but still beta — revisit at its 1.0); mypy (mature fallback). pyright not pursued due to Node JS requirement.
5. CI runner. With uv, run tools via `uv run … --output-format=github` for inline annotations (resolves the prek-vs-native tradeoff from #299); keep .pre-commit-config.yaml for local hooks.

### Open decisions

- Formatting: adopt ruff format despite the #298 diff (reconciling the .flake8 E226/E201/E241 matrix-spacing ignores), or skip indefinitely?
- Type-checker confirmation: pyrefly now / ty-when-stable / mypy — agree?

jonbiemond:
I ran ty and pyrefly over the codebase.

```bash
> ty check Pynite/
[...]

PR #333, opened 2026-08-05
Migrate packaging and CI to uv

tokito-99:
## Summary

Implements Phase 1 of #327 by migrating Pynite's packaging and environment management to `pyproject.toml` and uv.

This PR keeps the existing linting and testing behavior intact. The Ruff migration and type-checking work will be the next phases. 

## Changes

- replace `setup.py` with PEP 621 project metadata in `pyproject.toml`
- move runtime dependencies and published extras into `pyproject.toml`
- add a development dependency group for pytest, coverage, and Flake8
- replace `requirements.txt` with a committed `uv.lock`
- migrate the build-and-test workflow from ad-hoc pip installs to:
  - `setup-uv`
  - `uv sync --locked`
  - `uv run --locked`
- build the wheel and source distribution during CI
- migrate package publishing from setuptools/Twine to `uv build` and
  `uv publish`
- document the uv development workflow in `CONTRIBUTING.md`
- ignore uv's local `.venv` directory

The existing package metadata, optional dependency extras, supported Python versions, and bundled package data have been preserved.

## Verification

- [x] `uv lock --check`
- [x] `uv sync --locked --extra all --dev`
- [x] critical Flake8 checks pass with 0 violations
- [x] full test suite passes: 182 tests
- [x] wheel and source distribution build successfully
- [x] package metadata and bundled HTML/CSS files verified
- [x] GitHub Actions workflow YAML validated


pyproject.toml
```diff
@@ -0,0 +1,72 @@
...
+[project.optional-dependencies]
+all = [
+    "ipython",
+    "ipywidgets",
+    "jinja2",
+    "pdfkit",
+    "pyvista[all,trame]>=0.43.4",
+    "trame-jupyter-extension",
+    "vtk>=9.3.0",
+]
```
jonbiemond, line 39: The `--all-groups` flag can be used to select all groups. Can we drop these?

CONTRIBUTING.md
```diff
@@ -14,12 +14,28 @@ If you would like to contribute to Pynite, please review the following guideline
...
 7. Pull requests for any active repository `Projects` will usually be given first priorty.
 
+## Development Setup
+
+Pynite uses [uv](https://docs.astral.sh/uv/) to manage development environments and locked
+dependencies. Install uv, then create the environment and install the project with its test tools
+and optional runtime dependencies:
+
+```bash
+uv sync --locked --extra all --dev
```
jonbiemond:
```suggestion
uv sync --locked --all-groups
```

.gitignore
```diff
@@ -18,3 +18,4 @@ gui.py
 wxTableBases.py
 .env
 apps/*
+.venv/
```
jonbiemond: I don't think this is necessary. `.venv/` as its own `.gitignore` file. And it's something personal to developer's machines, so should really be set in each devs personal config. However the same could be same about a lot of the lines in this file.

.github/workflows/build-and-test.yml
```diff
@@ -20,35 +17,33 @@ jobs:
...
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install flake8 pytest
-        python -m pip install vtk
-        python -m pip install matplotlib
-        python -m pip install coverage
-        python -m pip install .
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Install the project
+      run: uv sync --locked --extra all --dev
```
jonbiemond:
```suggestion
      run: uv sync --locked --all-groups
```

jonbiemond (changes requested):
This is awesome! Just a couple suggestions to reduce some duplication. Also, is there maybe a way to move the versions to some sort of variable so they don't have to be updated across both workflows?

pyproject.toml
jonbiemond: Hi @tokito-99, do we still need this group? My concern is that the `all` group duplicates the dependencies of the other groups so there is a risk that they fall out of sync. For example, if vtk version is updated in just one group instead of both.

pyproject.toml
```diff
@@ -0,0 +1,75 @@
...
+
+[dependency-groups]
+dev = [
+    "coverage",
+    "flake8",
+    "pytest",
+]
+
+[tool.uv]
+required-version = "==0.12.1"
```
jonbiemond: Do you know if this means that contributor's local versions of uv need to match? That would be an inconvenience and potentially conflict with other projects. I think we should leave pinning the version to just CI. If there's not an easy way to set it in just one spot, I feel like the duplication between the two workflows is acceptable.

.github/workflows/build-and-test.yml
```diff
@@ -20,35 +17,32 @@ jobs:
...
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install flake8 pytest
-        python -m pip install vtk
-        python -m pip install matplotlib
-        python -m pip install coverage
-        python -m pip install .
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Install the project
+      run: uv sync --locked --extra all --all-groups
```
jonbiemond, line 28: I think the `--extra all` flag is not needed. Unless I'm missing something? (both workflow files)

jonbiemond (changes requested):

pyproject.toml
tokito-99: You’re right. An exact tool.uv.required-version constraint applies to contributors running uv locally in the project. I removed it and restored the CI-only version pins in both workflows. The small duplication is preferable to imposing an exact local uv version.

.github/workflows/build-and-test.yml
tokito-99: `--extra all` is required because `--all-groups` only installs development groups. Without the `all` extra, VTK, PyVista, and IPython are missing, causing four test modules to fail during collection.

pyproject.toml
```diff
@@ -0,0 +1,72 @@
+[build-system]
+requires = ["setuptools>=77"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "PyNiteFEA"
+version = "3.0.0"
+description = "A simple elastic 3D structural finite element library for Python."
```
JWock82: Remove the word "elastic". With version 3.0.0, Pynite now solves inelastic problems.

JWock82 (approved):

PR #338, opened 2026-09-14
Ruff unused variables

ulgens:
Another incremental improvement with lint setup. Enables F841: https://docs.astral.sh/ruff/rules/unused-variable/


JWock82:
I'll have to look at this one closer. I'm noticing this deletes some unused variables that may be placeholders for future features.

PR #340, opened 2026-09-23
Reject zero-length distributed loads in add_member_dist_load

### LINKED ISSUE #339: add_member_dist_load accepts zero-length loads and can divide by zero
## Problem

`FEModel3D.add_member_dist_load()` accepts equal `x1` and `x2` values. This creates a distributed load with zero length.

For a valid member, this input is accepted:

```python
model.add_member_dist_load("M1", "Fy", -1.0, -2.0, x1=5.0, x2=5.0)
```

During analysis, the member segmentation calculations divide by `(x2 - x1)`, so this invalid load can cause a division-by-zero failure instead of producing a clear input error.

## Impact

A malformed load definition is accepted at model construction time and fails later during analysis, making the source of the problem difficult to find.

## Acceptance criteria

- Reject a distributed load whose supplied interval is not strictly positive before analysis begins.
- Raise a clear `ValueError` when `x1 == x2`.
- Add a regression test showing the invalid load is rejected at input time.
- Preserve valid distributed-load behavior.

Ayberkrk:
Pynite/FEModel3D.py, add_member_dist_load(). The method accepts any x1/x2 pair without checking that the interval has positive length. Member3D.py then divides by (x2 - x1) multiple times during segmentation, so a load with x1 == x2, or x2 < x1, only fails once analyze() is called, with a bare ZeroDivisionError that gives no indication which load caused it.

Reproduced with a 5m beam and add_member_dist_load('M1', 'Fy', -1.0, -2.0, x1=5.0, x2=5.0): add_member_dist_load() accepts the load, and calling analyze() raises ZeroDivisionError: division by zero from deep inside the segmentation code.

Fix adds a check right after x1/x2 are resolved to their defaults, raising ValueError immediately if end <= start. This covers both the equal value case and a reversed interval, and it surfaces the problem at model construction time instead of mid-analysis.

Added test_dist_load_zero_length_rejected to Testing/test_loads.py, checking both x1==x2 and x2<x1 raise ValueError, and that a normal interval is still accepted. Verified it fails on the pre-fix code (ZeroDivisionError instead of ValueError) and passes with the fix.

Ran the full test suite in a clean venv. 183 passed, 18 subtests passed, excluding test_pushover.py which fails to even collect on this machine because matplotlib's TkAgg backend needs a system Tk installation this Python build doesn't have (ModuleNotFoundError: _tkinter), unrelated to this change.

Fixes #339


JWock82:
Thanks for this PR!

